PR #24, opened 2016-03-08
Refactor diff algorithms

martinal:
Work on streamlining the highlevel diff algorithms.


bollwyvl:
Looking forward to checking this out!

PR #26, opened 2016-03-08
Add git difftool prototype

minrk:
Register with:

    git nbdifftool config --enable [--global]

Use with:

    git difftool [<commit> [<commit>]]


bollwyvl:
Very cool!

PR #41, opened 2016-04-21
prettydiff fallback

bollwyvl:
> Congrats on inclusion in `jupyter`! :cake:

On a vanilla Windows install, `diff` is not available, so `test_present_string_diff` will always fail.

This adds a fallback based on stdlib `difflib.unified_diff`, and accepts either unified or diff-default.

Another approach would be to always use `diff -u` (and keep the fallback) but just always assume unified diff.


bollwyvl:
Huzzah!

PR #44, opened 2016-05-03
Git diff driver

vidartf:
Adds a notebook diff driver for git, which produces a diff for notebooks only. This can be used by `git diff` (as compared to `git difftool`), and relies on filename filtering in `.gitattributes` to be supplied the right files. The `.gitattributes`need to be configured for each repo, but the line included in this PR's `.gitattributes` file should be sufficient (`*.ipynb	diff=jupyternotebook`).

Also added a test that mocks a call by git to the diff driver, comparing the output. Currently it has to use coloring characters to match the captured output, but I'm not sure if this is correct/expected for all setups?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.ipynb	diff=jupyternotebook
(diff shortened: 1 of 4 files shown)


minrk:
Nice!

PR #49, opened 2016-07-07
Webapp w/difftool [WIP]

vidartf:
Creates a webapp that can either be launched as a normal server with an interface for diffing files (rather crude interface currently, with only two simple text boxes for serverside filename entry), or launched as a a one-shot server difftool from git.

#### Main things in this PR:
 - Setup for using webapp as difftool
 - Processing diff returned from server into a format for usable presentation. 
 - Logic to determine what should be show as code, direct JSON (e.g. metadata), or rendered MIME types.
 - Reimplements codemirror's mergeview such that it can better work in a future mergetool (it also solves some scrolling issues). This also allows the highlighting to be based on the Notebook diff supplied by the server, instead of relying on a simple diff of the stringified JSON.
 - Lays out processed diff in a series of side-by-side mergeviews
 
#### TODO (before merging?):
 - [ ] Better tests coverage
 - [ ] Ensure setup/packaging etc. is set up correctly
 - [ ] Documentation: Setup an auto API doc system for webapp + document webapp in sphinx

Should probably be squashed before merging, as commit history is a bit of a mess :cactus:


minrk:
Thanks!

If you want to cleanup the commits into representative chunks, feel free to do so and rebase. Alternately, we can merge with Squash & Merge, making this all a single commit (it's a lot of work for one commit, though!).

vidartf:
Sure, I'll do a backup of the history in a separate branch and do a soft nuke of this one (reset --soft + recommit) with hopefully somewhat descriptive commit messages.

vidartf:
Alright! Cleaned up the commits so that they follow a bottom-up build up of the code. Hopefully this will allow for bite-sized diffs for eased reviewing :smile:

vidartf:
PS: Commits also feature actual messages beyond the title, that are meant to be helpful :scream:

nbdime/webapp/nbdimeserver.py
```diff
@@ -173,7 +212,7 @@ def build_arg_parser():
     """
     description = 'Web interface for Nbdime.'
     parser = ArgumentParser(description=description)
-    parser.add_argument('-p', '--port',
+    parser.add_argument('-p', '--port', default="8888",
```
minrk: Default should be an integer, not string of integer. Then argparse will do the right type casting.

PR #50, opened 2016-07-07
Mimetype diff [WIP]

vidartf:
Implements a separate diff routine for outputs / mime bundles. This prevents large MIME data fields from being processed using character based diff, which scales horribly for large data strings. Now they are either processed as single blobs, or with a line-based diff.

This PR also implements a naive line-based diff algorithm (simply treats it as a list of strings), which can hopefully be improved in the future. Other fields/strings are also likely to benefit from using this instead of the character based diff, but I would wait with that discussion until @martinal is back in business.

#### TODO:
 - [ ] Tests
 - [ ] Determine which MIME types to treat as multiline strings, and which to treat as single blobs. Currently only 'text/*' MIME types are treated as strings.


nbdime/diffing/notebooks.py
```diff
@@ -133,23 +108,69 @@ def compare_output_data(x, y):
...
+    for key in sorted(akeys - bkeys):
+        di.remove(key)
+
+    # Handle values for keys in both a and b
+    for key in sorted(akeys & bkeys):
+        avalue = a[key]
+        bvalue = b[key]
+
+        # FIXME: Use linebased diff of some types of outputs:
+        # if key.startswith('text/') or key in _non_text_split_mimes = {
```
minrk: Maybe put a collection of mime-type patterns to test as a module global, so it's easier to add to. What you have here is a reasonable starting point.

Issue #51, opened 2016-07-13
build webapp in setup.py


minrk:
and make sure it's included in install

martinal:
On my setup with a virtualenv with npm installed via 'pip install nodeenv', setup.py doesn't build npm packages in nbdime/webapp and nbdime-web properly.

PR #65, opened 2016-07-16
Added a command to setup.py to run npm install and build

gahjelle:
See Issue #51.

I have added a command for running `npm install` and `npm build` during setup.py. The command is set to be run automatically on `build`, `develop`, and `sdist`. The code is mainly inspired by the similar code for jupyterlab, https://github.com/jupyter/jupyterlab/blob/master/setup.py

For now, the `npm install` checks for the `node_modules` directory, and runs if that directory does not exist, while `npm build` runs every time. There might be better ways to trigger this (I am not very familiar with node)?

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,5 @@ Install the [codecov browser extension](https://github.com/codecov/browser-exten
 
     git clone https://github.com/jupyter/nbdime
     cd nbdime
-    # for web-based diff tool:
-      cd nbdime/webapp
-      npm install && npm run build
-      cd ../..
     pip install .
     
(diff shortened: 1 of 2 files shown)


minrk:
@gahjelle thanks! This is a good start, and an improvement on what we had before.

> For now, the npm install checks for the node_modules directory, and runs if that directory does not exist, while npm build runs every time.

`npm install` should perhaps check if `node_modules` is newer than `package.json`. And by the time we release, `npm run build` should not run when installing from a release, as those will include the built js.

Issue #60, opened 2016-07-15
Markdown should be rendered


minrk:
including mathjax, etc.

While we don't have rich rendered markdown diff, we can still show source diffs there,
but at least unchanged markdown should be rendered.

vidartf:
:+1:

vidartf:
Fixed in #67, so can be closed.

PR #67, opened 2016-07-25
Render unchanged markdown

vidartf:
Ref #60. Not sure if all the dependecies for MathJax is included


.gitignore
```diff
@@ -18,3 +18,9 @@ nbdime/webapp/build/
 nbdime/webapp/test/build/
 
 nbdime/webapp/static/*bundle.js*
+nbdime/webapp/static/fa396483b7ade73b4727bad6aa027a35.html
```
minrk: These could presumably be blobs, right?

vidartf:
Refactored build system to use typescript loaders in webpack. This helps simplify the build process, and gives typescript sourcemaps for debugging. Also separates built static content into a separate folder for easy git-ignore.

PR #68, opened 2016-07-25
Handle empty root metadata diff

vidartf:
Previously this would fail the assert in `createPatchDiffModel()` for empty diffs in the assert.


minrk:
Needs rebase after #67, but 👍

vidartf:
:recycle: rebased

Issue #74, opened 2016-07-27
Use object to describe merge descisions


vidartf:
Motivated by a need to communicate the merge results/conflicts to the mergetool, I would propose to have the merge algorithm build a structure which defines its decisions. A gist outlining a proposal for such a format can be found here: https://gist.github.com/vidartf/5ff461d224583772ff799732956d8f07

This decision structure could then be applied to the base directly to produce the merged file or sent to the merge tool.

Advantages include:
 - Mergetool gets all the information it needs:
    - Which diffs are in conflict with each others.
    - Can easily show local/remote diffs aligned with merged output.
    - Can show the source of all difference in merged vs. base (i.e. color highlighting to say "this change comes from local, and this change comes from remote", which is good for context).
    - Has the original diffs easily accessible if the user wants to overrule/reset default decisions.
 - In the merge routine, the decision making is separated from how those decisions are applied to the notebook (e.g. inserting conflict markers). This should also help compartmentalize the code for easier maintenance and unit testing.

Input would be very welcome on:
 - Are there any cases where this will break down / cause inefficiencies?
 - Should there be any changes to the proposed format for describing the decisions?

martinal:
You convinced me of the general concept yesterday and I'm still with you.

Some input:

For the simple case, action could also be "either" when local and remote diffs are equal. Not important for the patching, but useful for the display?

I believe your actions list for the resolved case covers all relevant cases from the strategies list in nbdime/merging/notebooks.py.

I'm not sure if we need both type and action, maybe they can be combined:
- type=simple corresponds to action="local" or "remote" or "either"
- type=conflict corresponds to action=None (or maybe use "conflict" or "undecided")

martinal:
Also as we talked about yesterday, which may not be clear for other readers here, the merge algorithm produces this merge data structure but does not apply it in case of conflict (i.e. if any item here has action=None) but rather sends the base document and this structure for the GUI to apply. This seems to be a clean separation of concerns, and we avoid doing some patching here, some patching there.

martinal:
Basically when this structure is passed around, it represents different things:

- from the initial merge it represents the safe easy to resolve actions
- from the autoresolve stage it represents the opinionated resolutions
- when the gui gets this it can treat it as the recommendations to present to the user
- as the user works with the gui, this structure represents the resolutions the user chooses

minrk:
[...]

PR #75, opened 2016-07-28
Add a JSON schema describing merge decision format

vidartf:
See #74 for discussion about the format.

In addition to describing the format, it should be useful for unit testing.

### DIFF
--- a/nbdime/merge_format.schema.json
+++ b/nbdime/merge_format.schema.json
@@ -29,7 +29,7 @@
 
         "action": {
           "description": "How the merge was resolved, or for conflicted entries, the \"best\" initial suggestion for resolution",
-          "enum": ["local", "remote", "base", "clear", "custom"]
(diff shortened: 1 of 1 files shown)


nbdime/merge_format.schema.json
```diff
@@ -0,0 +1,144 @@
...
+        },
+
+        "conflict": {
+          "type": "boolean",
+          "description": "Whether the entry is conflicted or not"
+        },
+
+        "action": {
+          "description": "How the merge was resolved, or for conflicted entries, the \"best\" initial suggestion for resolution",
+          "enum": ["local", "remote", "base", "clear", "custom"]
```
martinal: What about "local_then_remote" and "remote_then_local"? Or would you cover that by combining the diffs into "custom_diff"?

And "either"? Although that's recoverable by checking if local_diff == remote_diff.

vidartf: I add all three in a new commit.

vidartf:
Schema updated as discussed in #74.

PR #76, opened 2016-08-04
Merge decisions

vidartf:
Implements merging using the merge decision format described in #74 (see also #75).

### Todo:
 - [x] Have autoresolve work with the new format.
 - [x] Update methods/tests that want both conflicts and merged output to use the new interface.
 - [x] Ensure parity with old method.
 - [x] Make new test / update old ones to test decision making and application separately.

Issue #33, opened 2016-04-15
Add wrapper script for git mergetool configuration


martinal:
For git mergetool to trigger nbmerge only for ipynb files, it seems we need to make a wrapper script that is invoked for all filetypes, checks the filetype, and forwards to another tool if it's not ours to merge.

A suggestion that could work with some modifications is here:
https://gist.github.com/elsevers/11349227

minrk:
This is what we already do in the difftool, so I think the same approach should work.

martinal:
Can we close this now?

vidartf:
I think we can say that it has been superseded by #88.

PR #77, opened 2016-08-18
Webapp mergetool

vidartf:
This PR implements a git mergedriver, and an initial, workable web-based mergetool. The mergetool still has some issues, and has not been tested against highly complex merges yet, but it could benefit from somebody else looking at it, and find some obvious missing cases.

To test the merge tool manually, use the console entry point `git-nbwebmergetool`. To test in a repo, run:

```
git-nbmergedriver config --enable [--global]
git-nbmergetool config --enable [--global]
```
This will first enable the merge driver (if `--global` is not used, you should be in the repo root where you want the driver to handle notebooks, so that it can find `.gitattributes`). Next it will enable the merge tool. Note that this will set the mergetool as the default, which will then only work for notebooks until #33 is implemented. You can manually change your tool back to your default after, and then use `git mergetool` with `-t`/`--tool=` to manually pick the tool when appropriate.

This PR is based on #76.

### TODO
 - [ ] Expand tests of webapp client-side code
 - [ ] Fix alignment on decision pickers on alignment offsets
 - [ ] Add widget for metadata conflict resolution on cell and notebook level
 - [ ] Add widget for outputs conflict resolution


martinal:
travis fails with
```
    /home/travis/build/jupyter/nbdime/nbdime/webapp/node_modules/jupyterlab/scripts/dedupe.js:7
    var version = childProcess.execSync('npm --version', { encoding: 'utf8' });
                               ^
    TypeError: Object #<Object> has no method 'execSync'
        at Object.<anonymous> (/home/travis/build/jupyter/nbdime/nbdime/webapp/node_modules/jupyterlab/scripts/dedupe.js:7:28)
```
Is this something that can be fixed?

nbdime/gitmergetool.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def show_merge(base, local, remote, merged):
+    """Run the mergetool
+
+    If we are merging a notebook, show the merge via nbmerge.
+    Otherwise, exit with error code.
+    """
+    # TODO: handle /dev/null (Windows equivalent?) for new or deleted files
+    if any([not f.endswith('.ipynb') for f in [base, local, remote, merged]]):
```
martinal: Tip: `any([x for x in mylist])` will build the list first and then run `any` on it. Instead, use a generator by dropping `[]`, i.e.:  `any(x for x in mylist)`  to let `any` abort on first True value. Not important here, but slightly less typing and improves performance.

nbdime/merging/decisions.py
```diff
@@ -0,0 +1,462 @@
...
+    Returns the new decision.
+
+    Raises a ValueErorr if a decision can not be created at the lower level.
+    """
+    diffs = [decision.local_diff, decision.remote_diff]
+    if decision.action == "custom":
+        diffs.append(decision.custom_diff)
+    popped = _pop_path(diffs)
+    if popped is None:
+        raise ValueError("Cannot pop patch decision for: " + str(decision))
```
martinal: Seems like you're using exceptions for expected program flow, may I suggest `return None` instead? (see below)

nbdime/merging/decisions.py
```diff
@@ -0,0 +1,462 @@
...
+
+    Calls `pop_patch_decision` recursively until it is at the lowest possible
+    decision level.
+
+    If the decision is already at the lowest level, it returns the original
+    decision.
+    """
+    try:
+        while 1:
+            decision = pop_patch_decision(decision)
```
martinal: Instead of `try/except` + `while 1:`, why not do `while decision is not None:`? (see above).

nbdime/gitmergedriver.py
```diff
@@ -31,6 +32,29 @@ def enable(global_=False):
     check_call(cmd + ['merge.jupyternotebook.driver', 'git-nbmergedriver merge %O %A %B %L %P'])
     check_call(cmd + ['merge.jupyternotebook.name', 'jupyter notebook merge driver'])
 
+    if global_:
+        try:
+            bpath = check_output(['git', 'config', '--global', 'core.attributesfile'])
```
martinal: `check_output` is not imported.

martinal:
@minrk did you take a look at the npm issue?

minrk:
Yes, I just opened https://github.com/vidartf/nbdime/pull/1 to install nodejs 4, which results in [passing tests](https://travis-ci.org/minrk/nbdime/builds/158741661).

martinal:
One pytest.raises type test case fails after you refactored the raising code.

vidartf:
Thanks, I'll start looking at splitting up this PR now.

vidartf:
Just rebased this. It is not a 100% polished yet with regards to the todo list, but the changes are so substantial at this point that I would argue for a merge and iterate approach.

martinal:
Agree, lets merge and get back to smaller PRs.

PR #80, opened 2016-09-23
[WIP] Webapp coverage

vidartf:
Adds unittest coverage from webapp.


vidartf:
Doesn't work currently, but a planned refactor of npm projects should fix this. Will resubmit this after that is in place.

stevepeak:
Did your node tests output a coverage report? The uploader should have discovered it.

I suggest you use the Bash uploader because the [flagging feature](http://docs.codecov.io/docs/flags).

```
# run python tests
py.test ...
bash <(curl -s https://codecov.io/bash) -F python

# run npm tests
npm tests
bash <(curl -s https://codecov.io/bash) -F javascript
```

This way you can issolate metrics from python and javascript.

vidartf:
@stevepeak : The problem was with remapping the coverage onto the original Typescript source, thanks though!

PR #82, opened 2016-10-03
Cleanup: CLI, setup and docs

vidartf:
Several areas were cleaned up: 

 - Refactors CLI arguments specification, cleaning up several issues related to them in the process.
 - Fixes merge format JSON schema + adds JSON schema for diff format.
 - Ensures setup.py runs correct npm commands after previous refactorization.
 - Updates dependencies (setup.py, requirements.txt, docs).
 - Add / improve docs.

### DIFF
--- a/nbdime/args.py
+++ b/nbdime/args.py
@@ -1,0 +1,89 @@
+# coding: utf-8
+
+# Copyright (c) IPython Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+from ._version import __version__
+
+
+def add_generic_args(parser):
+    """Adds a set of arguments common to all nbdime commands.
+    """
+    parser.add_argument(
+        '--version',
+        action="version",
+        version="%(prog)s " + __version__)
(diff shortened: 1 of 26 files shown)


nbdime/args.py
```diff
@@ -0,0 +1,95 @@
...
+    ))
+    parser.add_argument(
+        '-p', '--port',
+        default=default_port,
+        type=int,
+        help=port_help)
+    cwd = os.path.abspath(os.path.curdir)
+    parser.add_argument(
+        '-w', '--workdirectory',
+        default=cwd,  # TODO: Are there any security implications of doing this?
```
martinal: Possibly, it does expose cwd. But when starting `jupyter notebook`, cwd is the work directory by default so I think that  is precedence? @minrk

vidartf: This todo came along with a cut and paste, so I'm not the one asking the question. As I am not certain about the answer though, I left it untouched.

minrk: There aren't, starting in CWD is safe and common

PR #83, opened 2016-10-04
Consistency fixes in CLI tools

martinal:
… exit.

Also enable help text for all apps.

Canonical pattern:

    def main(args=None):
        if args is None:
            args = sys.argv[1:]
        ...
        return returncode

Outside of `__main__`, there's no need to have `if __name__ ==
"__main__"`.

With the addition of __main__, you can now do:

  python -m nbdime nbdiff <args>

instead of

  nbdiff <args>

which may be convenient to test with different python versions.

There is still some confusion about where --base --local --remote
are defined, I tried fixing a failure one place but the args.py utils
are not used consistently.

### DIFF
--- a/nbdime/__main__.py
+++ b/nbdime/__main__.py
@@ -1,0 +1,38 @@
+# coding: utf-8
+
+# Copyright (c) IPython Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+from __future__ import unicode_literals
+from __future__ import print_function
+
+import sys
+
+
+def main_dispatch(args=None):
+    if args is None:
+        args = sys.argv[1:]
+    cmd = args[0]
+    args = args[1:]
+
+    if cmd == "nbdiff":
+        from nbdime.nbdiffapp import main
+    elif cmd == "nbmerge":
+        from nbdime.nbmergeapp import main
+    elif cmd == "nbpatch":
+        from nbdime.nbpatchapp import main
+    elif cmd == "nbdiff-web":
+        from nbdime.webapp.nbdiffweb import main
+    elif cmd == "nbmerge-web":
+        from nbdime.webapp.nbmergeweb import main
+    else:
+        print("Invalid command '%s', expecting one"
+              "of diff, merge, or patch." % (cmd,))
+        return -1
+
+    return main(args)
+
+
+if __name__ == "__main__":
+    # This is triggered by "python -m nbdime <args>"
+    sys.exit(main_dispatch())
--- a/nbdime/__main__.py
+++ b/nbdime/__main__.py
@@ -26,9 +26,9 @@ def main_dispatch(args=None):
     elif cmd == "nbmerge-web":
         from nbdime.webapp.nbmergeweb import main
     else:
-        print("Invalid command '%s', expecting one"
-              "of diff, merge, or patch." % (cmd,))
-        return -1
+        print("Invalid command '%s', expecting one of:\n"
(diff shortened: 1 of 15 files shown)


martinal:
@vidartf The test failure here, what is it intended to check for?

vidartf:
I would keep the `if __name__ == "__main__":` check, as that allows `python -m nbdime.nbdiffapp` to work (we could even rename "nbdiffapp" to "nbdiff". This is a pretty standard execution pattern for apps. Having `python -m nbdime nbdiff` work is a bonus :)

The test failure is a check for a call to `sys.exit`, as the function was previously expected to exit. Can switch to simply checking return value.

nbdime/args.py
```diff
@@ -65,7 +66,6 @@ def add_diff_args(parser):
     #parser.add_argument('-d', '--diff-strategy',
     #                    default="default", choices=("foo", "bar"),
     #                    help="specify the diff strategy to use.")
-    pass
```
vidartf: To me `pass` represents "this function intentionally left empty", and keeps linters from complaining about empty functions.

nbdime/__main__.py
```diff
@@ -0,0 +1,38 @@
...
+        from nbdime.nbmergeapp import main
+    elif cmd == "nbpatch":
+        from nbdime.nbpatchapp import main
+    elif cmd == "nbdiff-web":
+        from nbdime.webapp.nbdiffweb import main
+    elif cmd == "nbmerge-web":
+        from nbdime.webapp.nbmergeweb import main
+    else:
+        print("Invalid command '%s', expecting one"
+              "of diff, merge, or patch." % (cmd,))
```
vidartf: This help text doesn't match the actual commands.

nbdime/__main__.py
```diff
@@ -0,0 +1,38 @@
...
+        return -1
```
vidartf: Everywhere else we use `return 1`. Shouldn't really matter, though (?)

nbdime/webapp/nbdimeserver.py
```diff
@@ -257,6 +257,8 @@ def _build_arg_parser():
     return parser
 
 
-if __name__ == "__main__":
-    arguments = _build_arg_parser().parse_args()
-    main(port=arguments.port, cwd=arguments.workdirectory)
```
vidartf: I would keep this as:
```python
if __name__ == "__main__":
    main()
```
so that the file can still be executed directly (`python nbdimeserver.py`). Is there any reason why we do not want this behavior? The `nd*-web` also pop open a browser and need a set of file names, while direct execution allows for a general server.

nbdime/webapp/nbmergeweb.py
```diff
@@ -21,20 +22,19 @@
...
+        )
     add_generic_args(parser)
     add_diff_args(parser)
     add_merge_args(parser)
     add_web_args(parser, 0)
-    parser.add_argument(
-        '-o', '--output',
-        default=None,
-        help="if supplied, the merged notebook is written "
-             "to this file. Otherwise it cannot be saved.")
```
vidartf: If you remove this, the merge application cannot save its result. Was this intended?

nbdime/__main__.py
martinal: 1 is more standard.

nbdime/args.py
martinal: I had added some lines there in an intermediate state. Will add back pass.

nbdime/webapp/nbmergeweb.py
martinal: It's not used in main below?

PR #90, opened 2016-10-10
Stricter typing, and transition to TS 2.0

vidartf:
All changes in this PR is to the web code.

It transitions the code to Typescript 2.0, and applies the `noImplicitAny` compiler flag as well as the new `strictNullChecks` compiler flag.

Also renames some properties, and updates some dependencies.


vidartf:
While this diff is big, most of this is only typing information (i.e. it disappears when compiled) or fixes revealed by the strict checks. As such, this should be a low-risk merge unless someone has any reservations about switching to strict null checks.

Issue #64, opened 2016-07-16
Automerge fails for two divergent notebook with the same base


eco32i:
The (minmal) test case:

https://gist.github.com/eco32i/2f9877b8044c28ef8a5ca525342ec73e

With these three notebooks executing the merge command fails:

```
$ nbmerge test.ipynb test_A.ipynb test_B.ipynb -o test_merged
Traceback (most recent call last):
  File "/home/ilya/.venv/pydata3/bin/nbmerge", line 9, in <module>
    load_entry_point('nbdime', 'console_scripts', 'nbmerge')()
  File "/mnt/data/scipy2016/nbdime/nbdime/nbmergeapp.py", line 109, in main
    r = main_merge(args)
  File "/mnt/data/scipy2016/nbdime/nbdime/nbmergeapp.py", line 43, in main_merge
    m, lc, rc = merge_notebooks(b, l, r, args)
  File "/mnt/data/scipy2016/nbdime/nbdime/merging/notebooks.py", line 54, in merge_notebooks
    autoresolve_notebook_conflicts(merged, local_diffs, remote_diffs, args)
  File "/mnt/data/scipy2016/nbdime/nbdime/merging/notebooks.py", line 36, in autoresolve_notebook_conflicts
    autoresolve(merged, local_diffs, remote_diffs, strategies, "")
  File "/mnt/data/scipy2016/nbdime/nbdime/merging/autoresolve.py", line 346, in autoresolve
    return autoresolve_dicts(merged, local_diff, remote_diff, strategies, path)
  File "/mnt/data/scipy2016/nbdime/nbdime/merging/autoresolve.py", line 317, in autoresolve_dicts
    newvalue, ldi, rdi = autoresolve(value, le.diff, re.diff, strategies, subpath)
  File "/mnt/data/scipy2016/nbdime/nbdime/merging/autoresolve.py", line 348, in autoresolve
    return autoresolve_lists(merged, local_diff, remote_diff, strategies, path)
  File "/mnt/data/scipy2016/nbdime/nbdime/merging/autoresolve.py", line 234, in autoresolve_lists
    newvalue, ldi, rdi = autoresolve(merged[j], le.diff, re.diff, strategies, subpath)
  File "/mnt/data/scipy2016/nbdime/nbdime/merging/autoresolve.py", line 346, in autoresolve
    return autoresolve_dicts(merged, local_diff, remote_diff, strategies, path)
[...]

PR #95, opened 2016-10-17
Fix merge cell delete and patch

martinal:
Various minor fixes, most important fix in merge cell deletion and metadata patch on other side.

Partially fixes
https://github.com/jupyter/nbdime/issues/64
but the execution_count handling issue is still left.

### DIFF
--- a/nbdime/webapp/package.json
+++ b/nbdime/webapp/package.json
@@ -16,6 +16,7 @@
     "css-loader": "^0.25.0",
     "file-loader": "^0.9.0",
     "json-loader": "^0.5.4",
+    "node-watch": "^0.4.1",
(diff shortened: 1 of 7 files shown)


nbdime/webapp/package.json
```diff
@@ -16,6 +16,7 @@
     "css-loader": "^0.25.0",
     "file-loader": "^0.9.0",
     "json-loader": "^0.5.4",
+    "node-watch": "^0.4.1",
```
vidartf: After a quick google, I think this should probably be this package instead: https://www.npmjs.com/package/watch,
That is: `"watch": "^1.0.1",`.

PR #100, opened 2016-10-20
Adjust pretty-print

vidartf:
Adjust pretty-print use of `unified_diff` to have the output be more similar to that of git diff.

### DIFF
--- a/nbdime/merging/generic.py
+++ b/nbdime/merging/generic.py
@@ -388,9 +388,11 @@ def _merge_strings(base, local_diff, remote_diff,
         _merge_strings.recursion = True
         base = base.splitlines(True)
 
-        _merge_lists(
-            base, local_diff, remote_diff, path, decisions)
-        _merge_strings.recursion = False
+        try:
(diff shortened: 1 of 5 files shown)


vidartf:
Fixes a failing test when git was not present, but I suspect this test will still fail if `diff` is used.

nbdime/merging/generic.py
```diff
@@ -383,14 +383,16 @@ def _merge_strings(base, local_diff, remote_diff,
...
         _merge_strings.recursion = True
         base = base.splitlines(True)
 
-        _merge_lists(
-            base, local_diff, remote_diff, path, decisions)
+        try:
+            _merge_lists(
+                base, local_diff, remote_diff, path, decisions)
+        finally:
+            _merge_strings.recursion = False
```
martinal: Sure this is the right way to break recursion?

vidartf: It is not very pretty, but it works (right?). Suggestions for a better pattern are welcome.

PR #102, opened 2016-10-26
Use io.open with utf-8 encoding everywhere

martinal:
Fixes nbdiff on commandline with notebooks containing unicode chars.


martinal:
Note that the tests added here don't actually cover the reported issue because they don't run the command line diff. But I've confirmed manually this branch fixes running nbdiff on the two `unicode--{1,2}.ipynb` files added here.

PR #110, opened 2016-10-31
Refactor models

vidartf:
Splits the monolithic model files in nbdime-web, as they were becoming rather unwieldy. Other related minor changes as well, as outlined in commit messages.


vidartf:
Rebased onto #109, and now passes CI.

PR #114, opened 2016-11-04
Allow nbdime web interfaces to load URLs

vidartf:
Adds ability for diff and merge web interfaces to load notebooks from URLs.

Also includes some other minor fixes and improvements.


martinal:
Works with nbdiff-web, nbmerge-web doesn't work for me at all now.

PR #117, opened 2016-11-08
Add smart index page

vidartf:
Implement index page, which will either perform a diff or a merge depending on how many filenames has been entered into the form.

Also fixes state management for back/forward navigation.


martinal:
@vidartf will you fix the conflicts?

vidartf:
@martinal - Should be OK now.

PR #118, opened 2016-11-08
Add error notifications in web-views

vidartf:
Adds a custom exception type, and a global exception handler that will display those errors to the user.

Adds an npm dependcy to `alertify.js` for displaying notifications (DOM-based).


vidartf:
Preview:
nbdime_error_notification

PR #125, opened 2016-11-14
Split decisions on cell source according to chunks

vidartf:
One-sided merge decisions or two-sided agreement decisions can sometimes contribute to more than one chunk in a string merge. E.g. a patch on line 5 and on line 10 in the cell source would belong to one decision. This would mean that when the user selects e.g. "local" for the patch on line 10, it will also pick "local" for line 5, which is unwanted behavior.

With this change, the web-based mergetool will split these types of decisions such that each decision only contribute to one chunk in the cell source.


vidartf:
Note that the code in `decisionchunking.ts` is mostly just a translation to typescript of the code in `chunks.py`.

nbdime-web/src/chunking/decisionchunking.ts
```diff
@@ -0,0 +1,221 @@
...
+import {
+  hasEntries, valueIn
+} from '../common/util';
+
+
+export
+type MergeChunk = {baseStart: number, baseEnd: number, diffs: DiffCollection}
+
+
+function any(diffs: DiffCollection): boolean {
```
martinal: IMO "any" is a bit too generic, anyDiffs maybe?

nbdime-web/src/chunking/decisionchunking.ts
```diff
@@ -0,0 +1,221 @@
...
+    if (e.op === 'addrange') {
+      // Pass
+    } else if (e.op === 'removerange') {
+      k = j + e.length;
+      boundaries.add(k);
+    } else if (e.op === 'patch') {
+      k = j + 1;
+      boundaries.add(k);
+    }
+}
```
martinal: Confusing indentation of }.

nbdime-web/src/chunking/decisionchunking.ts
vidartf: Agreed.

vidartf:
Comments fixed, and tests passing.

Issue #122, opened 2016-11-11
Add --browser argument to cli web tools


martinal:

martinal:
Implemented here:

https://github.com/jupyter/nbdime/tree/add-browser-argument

However the current error handling hides the failure to pick a browser if the name is not found, need to fix that before merging.

PR #126, opened 2016-11-14
Add -b/--browser argument to web tools.

martinal:


martinal:
Fixes https://github.com/jupyter/nbdime/issues/122

nbdime/webapp/nbdifftool.py
```diff
@@ -59,11 +61,12 @@ def main(args=None):
     cwd = arguments.workdirectory
     base = arguments.base
     remote = arguments.remote
+    browsername = arguments.browser
     return run_server(
         port=port, cwd=cwd,
         closable=True,
         difftool_args=dict(base=base, remote=remote),
-        on_port=browse)
+        on_port=lambda port: browse(port, browsername))
```
vidartf: `on_port` is a callback for when the port has been established, so calling browse directly here will not work with random port.

PR #128, opened 2016-11-16
Refactor web widgets

vidartf:
Split the rather large widget.ts into several smaller files, similar to what was done form models in #110.


martinal:
This is just splitting files?

vidartf:
Yeah, and changing some import paths. If there is anything else included it is unintentional.

vidartf:
I might also have ended up removing some unused variables/imports though.

PR #133, opened 2016-11-18
Allow diffs on both sides when encountering a 'clear' decision.

martinal:
I don't know if this is the right fix, however looking at this check isolated it makes sense: if a 'clear' decision is encountered, there may exist local _and_ remote diffs that we have decided to ignore.


martinal:
This fixes issue where
```
cd <nbdimerepo>/scratch/mergetesting
nbmerge-web -o webmerged.ipynb v1.ipynb v2.ipynb v3.ipynb
```
resulted in blank page. With this fix it shows the error about missing execution_count handling instead.

vidartf:
I think this needs to be handled differently, as the current fix will cause the remote diff to be dropped in the following code. In order to display a correct merge to the user we need to retain both.

vidartf:
Submitted #142 which should supersede this.

martinal:
I'm confused: you just closed #142, did you mean to do that?

vidartf:
Yeah, I realized I needed to adjust some stuff on the branch. Will reopen when I get it sorted.

Issue #113, opened 2016-11-03
Lots of build dependencies


ml31415:
npm, tornado, ... All I want is some nicer git diffs. Does it really need to have all these dependencies for that, no way to have that optional?

vidartf:
It should be possible to make the web tools as an optional feature (extras_require), or separate it out. Not sure what is the best way to achieve this.

vidartf:
I think the `--skip-npm` option for the install is as far as we're going to take this, together with the fact that releases (sdist/bdist) will not require `npm`. The tornado dependency will not be made optional, as it is so commonly required by other notebook tools as well.

PR #137, opened 2016-11-21
Add setup.py option --skip-npm

martinal:
Usage:

  pip install . --install-option="--skip-npm"

Addresses https://github.com/jupyter/nbdime/issues/113

Maybe there's a more standard way of doing this but this works.
Also nice while working only on python parts as the npm install is really slow.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,7 @@ from distutils.core import setup
 from distutils.cmd import Command
 from distutils.command.build import build
 from distutils.command.sdist import sdist
+from distutils.spawn import find_executable
 
 pjoin = os.path.join
 here = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -49,6 +50,17 @@
 ]
 
 
+if "--skip-npm" in sys.argv:
+    print("Skipping install of webtools as requested.")
+    skip_npm = True
+    sys.argv.remove("--skip-npm")
+elif not find_executable("npm"):
+    print("No npm found, skipping install of webtools.")
+    skip_npm = True
```
minrk: We should be careful about this - sdist should fail if npm is unavailable. Simiarly, install should fail if npm is unavailable and built js isn't present.

minrk: Thanks

PR #139, opened 2016-11-21
Improve  `attachments` cell field handling

vidartf:
Improves support for the `attachments` field in the python code (http://nbformat.readthedocs.io/en/latest/format_description.html#cell-attachments). Includes changes to the `prettyprint` code in order to detect attachments fields and apply base64 snipping to its mime data.

Support in the web-code will be added in a separate PR.

### DIFF
--- a/nbdime/diffing/notebooks.py
+++ b/nbdime/diffing/notebooks.py
@@ -144,7 +144,8 @@ def diff_single_outputs(a, b, path="/cells/*/output/*",
 _split_mimes = ('text/', 'image/svg+xml', 'application/javascript', 'application/json')
 
 
-def diff_mime_bundle(a, b, path=None):
+def diff_mime_bundle(a, b, path=None,
+                     predicates=None, differs=None):
     di = MappingDiffBuilder()
 
     akeys = set(a.keys())
@@ -194,6 +195,8 @@ notebook_differs = defaultdict(lambda: diff, {
     "/cells/*": diff,
     "/cells/*/outputs": diff_sequence_multilevel,
     "/cells/*/outputs/*": diff_single_outputs,
+    "/cells/*/attachments": diff_sequence_multilevel,
+    "/cells/*/attachments/*": diff_mime_bundle,
     })
 
 
--- a/nbdime/diffing/notebooks.py
+++ b/nbdime/diffing/notebooks.py
@@ -120,7 +120,6 @@ def diff_single_outputs(a, b, path="/cells/*/output/*",
     assert path == "/cells/*/outputs/*"
     assert a.output_type == b.output_type
 
-    # TODO: Handle output diffing with plugins? I.e. image diff, svg diff, json diff, etc.
     if a.output_type in ("execute_result", "display_data"):
         di = MappingDiffBuilder()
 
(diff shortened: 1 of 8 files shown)


nbdime/prettyprint.py
```diff
@@ -72,9 +72,9 @@ def present_dict_no_markup(prefix, d, exclude_keys=None):
                 pp.append(prefix + key + ':')
                 pp.extend(present_value(value_prefix, value))
             else:
-                pp.append(prefix + '%s: %s' % (key, value))
+                pp.extend(present_value(prefix + key + ': ', value))
         else:
-            pp.append(prefix + '%s: %s' % (key,  value))
+            pp.extend(present_value(prefix + key + ': ', value))
```
martinal: This is not what prefix is for, and besides I've rewritten the prettyprinting in another branch, so if you can drop this change it will make the merging cleaner later.

martinal: Nevermind. I'll incorporate the features you add below in this file in the new prettyprinting code so when the merge comes I can just drop your version and pick mine. You don't have to change this PR.

martinal:
If you wonder what's behind the tests failing here, it's that sequencematcher in difflib requires items in a list to be hashable. Strings and tuples are hashable, lists and dicts (and apparently 'slices') are not.

nbdime/prettyprint.py
vidartf: Ah, I see. Just note that this specific change was done to prevent base64 strings not being snipped properly.

vidartf:
Actually, the test failure was because I asked it to use a sequence matcher on an object (its trying to slice a dict).

nbdime/diffing/notebooks.py
```diff
@@ -194,6 +195,8 @@ def diff_mime_bundle(a, b, path=None):
     "/cells/*": diff,
     "/cells/*/outputs": diff_sequence_multilevel,
     "/cells/*/outputs/*": diff_single_outputs,
+    "/cells/*/attachments": diff_sequence_multilevel,
+    "/cells/*/attachments/*": diff_mime_bundle,
```
vidartf: These guys will not work as they should since `attachments` is a dict of named bundles so:
 - `diff_sequence_multilevel` doesn't make sense
 - `"/cells/*/attachments/*"` will never match, as currently star only matches list indices (I think?)

martinal: I suggest adding a diff_attachments, don't think it will be hard?

martinal: I suggest adding a diff_attachments, don't think it will be hard?

vidartf: Agreed, do you want me to push it to the branch of this PR?

martinal: Please do.

vidartf: I started working on it, and a major question appeared: If someone has renamed an attachment (identical or short edit of content, but different key), how can we represent this? For normal notebook content this is maybe not so relevant, but this would be likely for attachments. Should we introduce a `rename` diff op?

vidartf: (or more generally a `move` op, but a move op for arrays is likely to add more complexity)

martinal: I'm inclined to skip that at least for now, we should focus on polishing and making existing features robust now. Agree that a rename op would be much easier than an array move op though, but it still needs to be handled across quite a few places in the codebase.

vidartf: Pushed.

PR #140, opened 2016-11-21
Fix bug in merge for base=[y,z], local=[x,z], remote=[x,y].

martinal:


vidartf:
For future reference, the conclusion was that the removed code was part of an earlier attempt at solving addrange+removerange vs addrange+removerange conflicts (AR/AR in notation used in code), but is not needed as we currently only use it for AR/A conflicts.

PR #145, opened 2016-11-25
simplify conditions in setup.py

minrk:
There should be fewer different circumstances to handle:

- always use setuptools
- disable implicit bdist_egg (this is the main reason to avoid setuptools)
- wrap build_ext instead of build & develop (build_ext is called before both)

- simplify skip_npm handling: perform the skip in the npm-using commands, rather than changing what commands exist.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -210,7 +210,7 @@ class bdist_egg_disabled(bdist_egg):
 
 
 cmdclass = dict(
-    build_ext  = js_prerelease(build_ext),
+    build_ext = js_prerelease(build_ext),
     sdist  = js_prerelease(sdist, strict=True),
     jsdeps = combine_commands(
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -193,24 +199,27 @@ def run(self):
...
+    Prevents setup.py install performing setuptools' default easy_install,
+    which it should never ever do.
+    """
+    def run(self):
+        sys.exit("Aborting implicit building of eggs. Use `pip install .` to install from source.")
+
+
+cmdclass = dict(
+    build_ext  = js_prerelease(build_ext),
+    sdist  = js_prerelease(sdist, strict=True),
```
vidartf: Maybe add `bdist_wheel` with `strict=True` as well?

minrk: ✔️

PR #149, opened 2016-11-28
Add NOTSET choice for --log-level and set default to INFO.

martinal:
@minrk consider including this change or not, depending on your current log changes! Maybe the default should still be warning.


vidartf:
The default as `INFO` is already in. After looking into it, `NOTSET` doesn't really make sense from CLI (its end effect will be equal to `DEBUG`).

PR #150, opened 2016-11-28
Mergetool command

### LINKED ISSUE #88: Script/alias to run mergetool on notebooks
While git tools cannot selectively be run on certain file extensions, the following command is very useful, and we should maybe make a command for it?

```
git mergetool --tool=nbdimeweb *.ipynb
```

Which basically leverages the optional file field of the `git mergetool` command, and the fact that it accepts wildcards.

Suggestions for command name if we implement this?

vidartf:
- Adds `nbdime` as console entry point to main dispatch.
 - Renames main dispatch sub-commands (removes `nb` prefix).
 - Adds main dispatch sub-command (`mergetool`)

Fixes #88.

### DIFF
--- a/nbdime/__main__.py
+++ b/nbdime/__main__.py
@@ -7,6 +7,7 @@ from __future__ import unicode_literals
 from __future__ import print_function
 
 import sys
+COMMANDS = ["show", "diff", "merge", "patch", "diff-web", "merge-web"]
 
 
 def main_dispatch(args=None):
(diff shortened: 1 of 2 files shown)


vidartf:
Rebased after #151 .

nbdime/__main__.py
```diff
@@ -7,34 +7,46 @@
...
+    if cmd == "show":
         from nbdime.nbshowapp import main
-    elif cmd == "nbdiff":
+    elif cmd == "diff":
         from nbdime.nbdiffapp import main
-    elif cmd == "nbmerge":
+    elif cmd == "merge":
         from nbdime.nbmergeapp import main
-    elif cmd == "nbpatch":
+    elif cmd == "patch":
```
martinal: Maybe remove patch here. We're not currently planning on implementing fuzzy patching, so it's not very useful and doesn't behave as users would expect from outside diff/patch experience.

vidartf: Done!

PR #151, opened 2016-11-28
Update logging and entrypoints

minrk:
- remove non-web git diff/merge tools (and remove web from names of remaining diff/merge tools)
- consolidate init_logging calls
- trigger set_nbdime_log_level automatically with argparse Action
- expose `nbdime.log.info,error,etc.` log functions
- remove nbpatch entrypoint (leave code alone)


vidartf:
Looks ok!

PR #153, opened 2016-11-28
Merge strategies CLI

### LINKED ISSUE #19: Add merge strategy options
For example git merge has options like 'use ours'/'use theirs' to force conflict resolution, and we can add notebook-specific merge options like 'clean all notebook output', 'clean conflicting notebook output', 'join conflicting notebook output', and many more. We should sketch some use cases to figure out which options will be useful and try to keep the set small.

vidartf:
Refactor the CLI for merge strategies.
 - Limit applicable CLI merge strategies to "use-base", "use-local", "use-remote", "union",  or "inline".
 - Functions also accept "mergetool" as strategy.
 - Flag `--no-ignore-transients` : Disallow clearing/ignoring of transient data on conflicts.
 - Added `--input-strategy`/`--output-strategy`, that takes the same values as `--merge-strategy`, but will override strategy for source/outputs, respectively.

Fixes #19.

### TODO:
 - [x] Add tests for input/output strategies.
 - [x] Add tests for interplay between various flags and options (order of precedence).
 - [x] Document interface

### DIFF
--- a/nbdime/args.py
+++ b/nbdime/args.py
@@ -101,7 +101,7 @@ def add_merge_args(parser):
         dest='ignore_transients',
         action="store_false",
         default=True,
-        help="Allow deletion of transient data such as outputs and "
+        help="Disallow deletion of transient data such as outputs and "
              "execution counts in order to resolve conflicts.")
 
 
--- a/nbdime/args.py
+++ b/nbdime/args.py
@@ -78,16 +78,29 @@ def add_diff_args(parser):
 def add_merge_args(parser):
     """Adds a set of arguments for commands that perform merges.
     """
-    from .merging.notebooks import generic_conflict_strategies
+    from .merging.notebooks import cli_conflict_strategies
     parser.add_argument(
         '-m', '--merge-strategy',
-        default="mergetool",
-        choices=generic_conflict_strategies,
+        default="inline",
+        choices=cli_conflict_strategies,
         help="Specify the merge strategy to use.")
     parser.add_argument(
-        '-i', '--ignore-transients',
-        action="store_true",
-        default=False,
+        '--input-strategy',
+        default=None,
+        choices=cli_conflict_strategies,
+        help="Specify the merge strategy to use for inputs "
+             "(overrides 'merge-strategy' for inputs).")
+    parser.add_argument(
+        '--output-strategy',
+        default=None,
+        choices=cli_conflict_strategies,
+        help="Specify the merge strategy to use for outputs "
+             "(overrides 'merge-strategy' for outputs).")
+    parser.add_argument(
+        '-i', '--no-ignore-transients',
+        dest='ignore_transients',
+        action="store_false",
+        default=True,
         help="Allow deletion of transient data such as outputs and "
              "execution counts in order to resolve conflicts.")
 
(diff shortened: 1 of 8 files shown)


nbdime/args.py
```diff
@@ -78,17 +78,30 @@ def add_diff_args(parser):
...
+        help="Specify the merge strategy to use for inputs "
+             "(overrides 'merge-strategy' for inputs).")
+    parser.add_argument(
+        '--output-strategy',
+        default=None,
+        choices=cli_conflict_strategies,
+        help="Specify the merge strategy to use for outputs "
+             "(overrides 'merge-strategy' for outputs).")
+    parser.add_argument(
+        '-i', '--no-ignore-transients',
```
vidartf: I guess this short notation doesn't really make sense with a negative flag?

vidartf: Removed for now.

vidartf:
Should be read for review now.

nbdime/merging/autoresolve.py
```diff
@@ -414,12 +431,13 @@ def autoresolve_decision_on_dict(dec, base, sub, strategies):
     if strategy is not None:
         decs = strategy2action_dict(sub, le, re, strategy, subpath, dec)
     elif le.op == DiffOp.PATCH and re.op == DiffOp.PATCH:
+        assert False
```
martinal: Are we thinking alike if I say this will never happen because patches have previously been popped up to the common path?

nbdime/merging/autoresolve.py
```diff
@@ -474,7 +494,8 @@ def autoresolve_decision(base, dec, strategies):
     elif isinstance(sub, list):
         decs = autoresolve_decision_on_list(dec, base, sub, strategies)
     elif isinstance(sub, string_types):
-        raise NotImplementedError()
+        sub = sub.splitlines(True)
+        decs = autoresolve_decision_on_list(dec, base, sub, strategies)
```
martinal: Should this process the decisions back to a "string context" after resolution in "list context"? Or have you resolved that some other way?

nbdime/args.py
```diff
@@ -78,17 +78,30 @@ def add_diff_args(parser):
...
+             "(overrides 'merge-strategy' for inputs).")
+    parser.add_argument(
+        '--output-strategy',
+        default=None,
+        choices=cli_conflict_strategies,
+        help="Specify the merge strategy to use for outputs "
+             "(overrides 'merge-strategy' for outputs).")
+    parser.add_argument(
+        '--no-ignore-transients',
+        dest='ignore_transients',
```
martinal: Suggestion for alternative name: -s, --strict

nbdime/args.py
```diff
@@ -78,17 +78,30 @@ def add_diff_args(parser):
...
+        action="store_false",
+        default=True,
+        help="Disallow deletion of transient data such as outputs and "
```
martinal: This gives the impression that we delete outputs to resolve conflicts without this flag.

nbdime/merging/autoresolve.py
vidartf: Oh. This one kind of slipped through, but I included it to test that hypothesis, yes.

nbdime/args.py
vidartf: Sounds good to me, as negative flags are always a bit confusing.

nbdime/args.py
vidartf: (in which case we delete the entire cell, and not only the output)

PR #154, opened 2016-11-28
Update documentation for installation and index intro

willingc:
Pretty heavy edit of content in `index.rst` and `installation.rst` and minor edits to `conf.py`.

- `index.rst` mostly styling and grammar consistency for reader
- `installation.rst` updated installation instructions for virtualenv for Python 2 and Python 3

Do we want to revisit which versions of Python 2 and Python 3 are supported?

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -26,7 +26,7 @@ import recommonmark.parser
 # -- General configuration ------------------------------------------------
 
 # If your documentation needs a minimal Sphinx version, state it here.
-needs_sphinx = '1.3'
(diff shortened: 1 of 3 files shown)


martinal:
Looks good!

Great if someone more experienced with cross platform python/node.js installation can review recommendations for the pip/venv/nodeenv instructions, I only know that this works for me on Ubuntu 16.04.

Any particular changes in mind for the python versions?

minrk:
I think leaving the current versions is fine. I don't think there are any relevant differences between 3.3 and 3.4 for us here.

Thanks, @willingc!

PR #159, opened 2016-11-30
Edit docs 

willingc:
- Add doc-requirements.txt for local builds
- Edit all installation and usage docs
- Correct TODO directives to suppress the notes from displaying on RTD


minrk:
Great!

PR #161, opened 2016-11-30
remove some global nbdimeserver state

minrk:
attach things to app.settings

rather than module-global


nbdime/webapp/nbdimeserver.py
```diff
@@ -239,6 +240,7 @@ def make_app(**params):
     settings = {
         "static_path": static_path,
         "template_path": template_path,
+        "exit_code": 0,
```
vidartf: I would maybe not put this in `settings`, as `tornado` uses that object. Put it as an attribute?

minrk: done

PR #165, opened 2016-11-30
add screenshots for docs

minrk:


vidartf:
I'll merge for now, but if a native English speaker can consider my comment above for a separate PR, that would be nice :)

PR #155, opened 2016-11-29
Various fixes

martinal:
Introduces support for git merge-file and diff3 for rendering inline source conflicts.

Also contains a few smaller fixes and edits I've done while looking around the code.

Includes the log fix from @minrk.

Inlining of source and output conflicts do _not_ work as expected yet.

I'd like to have this branch merged now because it contains various useful small fixes, and will create a new PR for conflict inlining later.

PR #167, opened 2016-11-30
Partial cleanup after #155

vidartf:
Fixes part of the hanging issues after #155 + some random cleaning.

### DIFF
--- a/nbdime/merging/autoresolve.py
+++ b/nbdime/merging/autoresolve.py
@@ -6,7 +6,6 @@
 from __future__ import print_function, unicode_literals
 
 from six import string_types
-import sys
 import copy
 import logging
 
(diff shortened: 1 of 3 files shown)


nbdime/merging/autoresolve.py
```diff
@@ -190,6 +189,8 @@ def strategy2action_dict(resolved_base, le, re, strategy, path, dec):
         mval = max(bval, lval, rval)
         if bval == mval:
             return []
+        elif lval == mval == rval:
+            dec.action = "either"
```
minrk: Is there a difference between "either" and "local" when they are the same?

martinal: Not in the end result; either implies that local and remote are equal.

martinal:
Thanks for cleaning up after me ;)

nbdime/merging/autoresolve.py
vidartf: The presentation framework can use it, as it should be of interest to the user whether local work picked, or the changes agree.

PR #169, opened 2016-11-30
Fix autoresolve of addrange+patch vs addrange+patch conflicts

vidartf:
- Adds some test files which gives AP/AP conflicts (with either differing or identical patch ops).
 - Fixes handling of such conflicts.

Note that the implemented handling of such conflicts for strings is a temporary fix in wait of a more complete overhaul of string diff chunking.

### DIFF
--- a/nbdime/merging/generic.py
+++ b/nbdime/merging/generic.py
@@ -10,8 +10,10 @@ from six import string_types
 from .decisions import MergeDecisionBuilder
 from .chunks import make_merge_chunks
 from ..diffing import diff
-from ..diff_format import (DiffOp, as_dict_based_diff, op_patch, op_addrange)
+from ..diff_format import (
+    DiffOp, as_dict_based_diff, op_patch, op_addrange, op_removerange)
 from ..diffing.notebooks import notebook_predicates, notebook_differs
+from ..patching import patch
 from ..utils import star_path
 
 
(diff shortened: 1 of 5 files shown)


martinal:
Are you sure this should be autoresolved at all? It's not the end goal of autoresolve to get rid of _all_ conflicts, sometimes we just want to pass on to the mergetool for human inspection.

vidartf:
If the patches agree, I believe they can safely be resolved. For other cases, I believe decisions on cells make sense to recurse (while string decisions are kept conflicted).

martinal:
git merge-file and diff3 do not merge inserts at same location, even without patch:
```
$ cat a.txt
# this line is common
# here this other line is added
# a shared line at the end
$ cat b.txt
# this line is common
# with this line inserted
# a shared line at the end
$ cat c.txt 
# this line is common
# a shared line at the end
```
```
$ git merge-file -p a.txt c.txt  b.txt
# this line is common
<<<<<<< a.txt
# here this other line is added
=======
# with this line inserted
>>>>>>> b.txt
# a shared line at the end
```
```
$ diff3 -m a.txt c.txt  b.txt
# this line is common
<<<<<<< a.txt
# here this other line is added
||||||| c.txt
=======
# with this line inserted
>>>>>>> b.txt
# a shared line at the end
```

Issue #168, opened 2016-11-30
Doc link to jupyter notebook doesn't work


vidartf:
The link in this line of `index.rst`: 
```rst
**nbdime** provides tools for diffing and merging `Jupyter notebooks <jupyter-notebook>`_.
```
gives a 404 on RTD, in the repo, and locally.

willingc:
Fixed in #171

PR #171, opened 2016-11-30
Add captions and alt text to images in index doc

willingc:
I love the new screenshots :smile:

I've added accessibility alt-txt and captions to the images on the index page to provide the reader more context when first reading the doc.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -1,12 +1,14 @@
 nbdime -- diffing and merging of Jupyter Notebooks
 ==================================================
 
-.. image:: images/nbdiff-web.png
-
 Version: |version|
 
 **nbdime** provides tools for diffing and merging `Jupyter notebooks <jupyter-notebook>`_.
 
+.. figure:: images/nbdiff-web.png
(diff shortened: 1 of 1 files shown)


willingc:
Closes #168

minrk:
Great! A couple of questions:

- I see you switched from image to figure. Is there a reason for the change (so I can know the right thing to use in the future)?
- The screenshots also show up in cli.rst and merging.rst, were you going to update those, as well?

willingc:
With rst, you would use `figure` if you wish to add a caption. `image` doesn't offer that. Typically, I use `image` and only use `figure` when I wish to add a caption. [Reference](http://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#images-and-figures)

I wasn't sure if we wanted captions in `cli.rst` or `merging.rst` since the images reflected the subsection where in `index.rst` I felt the captions helped reinforce the text. Happy to add if desired.

willingc:
Probably would be helpful to add the alt-text to the images in `cli.rst` and `merging.rst`. I'll add that in a separate PR.

PR #174, opened 2016-12-02
fix args handling in merge driver, app

minrk:


nbdime/nbmergeapp.py
```diff
@@ -48,12 +48,12 @@ def main_merge(args):
     else:
         nbdime.log.debug("Merge completed successfully with no unresolvable conflicts.")
 
-    if args.decisions:
+    if decisions:
```
martinal: No, args.decisions is deliberate, it shouldn't print decisions without --decisions on the commandline.

nbdime/nbmergeapp.py
```diff
@@ -48,12 +48,12 @@ def main_merge(args):
...
         # Print merge decisions (including unconflicted)
         out = io.StringIO()
         pretty_print_merge_decisions(b, decisions, out=out)
         nbdime.log.warning("Conflicts:\n%s", out.getvalue())
-    elif mfn:
+    if mfn:
```
martinal: This was also deliberate, if --decisions are asked for, it doesn't print the merged notebook.

nbdime/nbmergeapp.py
minrk: ah, ok. Then there's a problem because `args.decisions` is undefined.

minrk:
Thanks for the clarification, @martinal! I figured out that the problem was merge args being added in the wrong place. I've moved them to the common merge_args location, so they are shared by the various merge apps.

minrk:
Oops, wrong again, it was just in the mergedriver. I've made the smaller fix. I'm working on another PR for greatly increased CLI coverage, but this should fix the mergedriver for now.

martinal:
Great, we need that.

PR #175, opened 2016-12-02
Fixes for inline merge strategies

martinal:
This includes the implementation of inline strategy for source based on our chunking, leading to excessive markers, which is a slightly better state than failure in case of multiple conflicts. To be improved later!

Adds nbmerge output-strategy option 'clear', not yet working properly.

Also various small changes.


vidartf:
Inline-source should work now, output strategy clear still needs work

vidartf:
I've pushed some commits for output clearing now as well. It works, but is maybe not a 100% (e.g. doesn't ignore transients). The names are also maybe not ideal (clear vs remove vs clear-all), but should be ok enough for a merge?

martinal:
Good job. I'm working on testing it, clear-all seems to work, remove as well, looking at a bug in output inlining. It's no doubt a big step forward, so I'm fine with merging now and looking into the bug for another PR.

PR #185, opened 2016-12-05
Use node 6 for travis

vidartf:
Should fix JS tests failing.


vidartf:
Passes all tests, should be ready to merge.

Issue #141, opened 2016-11-22
Update docs for next release


willingc:
@martinal @minrk Opening this as a checklist for any doc updates prior to the next release. Please add or update. Happy to pitch in before after Thanksgiving holiday here.

- [x] Review installation instructions in Sphinx docs
- [x] Review installation instructions in README
- [x] Add any new features/functionality to Sphinx docs
- [x] Add a changelog doc (Initial release yay!)

martinal:
Thanks! We're aiming for initial release next week, in addition to another iteration on the documentation there are some minor issues we'd like to fix first to make the merge tools more usable.

vidartf:
What remain here? Only changelog, or is there any other new features/functionality?

vidartf:
I guess we should remove difftool docs, as per #147?

vidartf:
Mergetool docs should probably include the shortcut `nbdime mergetool` = `git mergetool --tool=nbdimeweb *.ipynb`.

minrk:
I'd leave the difftool docs for now, while we don't have an implementation of #147. I've got a PR coming shortly finishing up the various TODO items, then we should be set.

minrk:
I don't think we should have too many ways to call these things. I would leave out the `nbdime mergetool` doc, and possibly remove the `nbdime` entrypoint. I think it's handy to have `python -m nbdime` for certain cases, but I don't think `nbdime nbdiff` is helpful, for instance.

martinal:
Agree on not having too many ways to call. Which cases are `python -m nbdime` useful for? If the only use is to allow running with specific python version, which can be useful for testing, then maybe we can just keep it but not document it?

minrk:
> If the only use is to allow running with specific python version, which can be useful for testing, then maybe we can just keep it but not document it?

That's the main one. I'm happy to have it, but not cover it in the sphinx docs.

vidartf:
I wanted to include the `nbdime mergetool` shortcut, as it is much easier to remember than `git mergetool --tool=nbdimeweb *.ipynb` (which is also a bit of a hassle to write).

minrk:
Ah, that makes sense. The rest don't as much, though. Plus, shouldn't the alias just be for `git mergetool --tool=nbdimeweb`, not including the paths?

vidartf:
[...]

PR #193, opened 2016-12-06
Init changelog for 0.1

minrk:
and get docs version from nbdime in conf.py

last task for #141

cc @willingc


willingc:
@minrk @martinal @vidartf Congrats on the release!

PR #194, opened 2016-12-06
Output merge improvements

martinal:
Tries to mark outputs better based on diffs instead of throwing all outputs between markers with output-strategy inline.

Also please review the changes in diff for attachments, these are untested.

### DIFF
--- a/nbdime/merging/autoresolve.py
+++ b/nbdime/merging/autoresolve.py
@@ -8,6 +8,7 @@ from __future__ import print_function, unicode_literals
 from six import string_types
 import copy
 import logging
+from itertools import chain
 
 import nbformat
 from nbformat import NotebookNode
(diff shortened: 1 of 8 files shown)


vidartf:
There should be some attachment notebooks in the test files folder, (hand crafted).

martinal:
Don't merge this, I'm working on fixes in the output compare predicates.

martinal:
Ready for review now. Includes an attempt at conflict handling for attachments for inline mode.

There's a bug in decision bundling which affects attachments, I'll submit an issue on that separately since attachments are experimental anyway.

nbdime/merging/autoresolve.py
```diff
@@ -562,6 +625,121 @@ def make_inline_source_decision(source, prefix, local_diff, remote_diff):
     )]
 
 
+def make_inline_attachments_decision(attachments, prefix, local_diff, remote_diff):
```
martinal: @vidartf here (mentioned in https://github.com/jupyter/nbdime/issues/197)

Issue #197, opened 2016-12-07
All merge strategies on non-leaf nodes need bundled decisions


martinal:
Including 'union', and 'record-conflict'.

Typical symptom:
```
    def patch_dict(obj, diff):
        newobj = {}
        deleted_keys = set()
    
        for e in diff:
            op = e.op
            key = e.key
            assert isinstance(key, string_types)
>           assert key not in newobj
E           AssertionError

nbdime/patching.py:95: AssertionError
```

We should consider rewriting the merge code to apply strategies earlier instead of creating decisions first and then collecting them again.

vidartf:
To me, the biggest open question is when transients checks should be applied. If that can be (correctly) included in bundling code, I think we can make our job easier.

martinal:
If we apply strategies early as I suggest there would be no need for bundling decisions afterwards.

Take a look at the inline conflicts handler for attachments in this PR:
https://github.com/jupyter/nbdime/pull/194
notice how it has to re-do basically the entire merge of the attachments dict?

In the original dict the diff on e.g. `/cells/3/attachments` should already be bundled in a patch.

I don't remember why we didn't do it that way initially, have I forgotten something important?

PR #198, opened 2016-12-07
Add merge tests and bundle decisions for record-conflict strategy

martinal:
Opening this WIP PR for the record and discussion.

This builds on https://github.com/jupyter/nbdime/pull/194, the new commits begin with "Add some new merge tests, already provoking a bug".

This started as an attempt to systematically add merge tests for the inline merge strategies.

I didn't get far before I hit this:
https://github.com/jupyter/nbdime/issues/197

The bundling of metadata conflicts done so far in this PR needs to deal with re-merging the non-conflicting changes from local_diff and remote_diff to produce the right result.

I believe we should apply the strategy handlers we've created at an eariler stage, in the first pass generic merge to produce the decisions right away. Note that all the make_foo_decisions handlers are take the base object and diffs as input, not decisions. Opinions?


martinal:
To clarify: This branch begins to (but does not complete) handle the record-conflicts strategy for metadata but stopped because of a bug in bundling. I'm currently refactoring and reworking the merge routines to get around the fundamental issue instead of adding more layers of complexity.

martinal:
I've pushed the current state of merge rework, not complete. Points left to do that I'm aware of are marked XXX and/or FIXME.

vidartf:
With added commits to #194 and subsequent merge, would you mind rebasing this for a clearer overview of what is done here and in previous?

martinal:
I've added comments on what's left to do, see top of generic.py, in case someone wants to attempt to continue. I'll continue on monday.

nbdime/merging/generic.py
```diff
@@ -738,34 +738,74 @@ def _merge_concurrent_inserts(base, ldiff, rdiff, path, decisions, strategies=No
...
+    #    insert [7] at 1 (base index)"
+    # instead we now have this which messes up the ordering:
+    #   "insert [7] at 1 (base index);
+    #    conflicting insert [2] vs [3] at 1 (base index)"
+    # perhaps change to this:
+    #   "insert [7] at key=1 (base index) lkey=1 rkey=1;
+    #    conflicting insert [2] vs [3] at key=1 (base index) lkey=0 rkey=0"
+    # then decisions can be sorted on (key,lkey) or (key,rkey) depending on chosen side.
+    # This test covers the behaviour:
+    # py.test -k test_shallow_merge_lists_insert_conflicted -s -vv
```
martinal: @vidartf something to discuss on Wednesday.

Note that this is related to how conflicts of inserts should be treated, and diff3 and git merge-file makes different choices:
```
$ more a.txt c.txt b.txt 
::::::::::::::
a.txt
::::::::::::::
1
4
3
::::::::::::::
c.txt
::::::::::::::
::::::::::::::
b.txt
::::::::::::::
1
2
3

$ diff3 -m a.txt c.txt b.txt 
<<<<<<< a.txt
1
4
3
||||||| c.txt
=======
1
2
3
>>>>>>> b.txt

$ git merge-file -p a.txt c.txt b.txt 
1
<<<<<<< a.txt
4
=======
2
>>>>>>> b.txt
3
```

Perhaps the easiest is to follow diff3 and ignore similarities between local and remote insertions.

martinal: (Sidenote: I had to remind myself why identical inserts are relevant: it can occur in large amounts if git branches are rebased too aggressively and later merged again.)

minrk:
@martinal I got your test passing. The issue was that consecutive decisions on the same key (`metadata.unconflicted` in this case) could result in consecutive patches, in this case:

```
- metadata.unconflicted.list_deleteitem: patch.removerange(1,1)
- metadata.unconflicted.list_deleteitem: patch.removerange(5,1)
```

Applying the `combine_patches` to the diffs will flatten these appropriately, and the patch will apply.

martinal:
Thanks! I suspected something like that, recognised the type of error.

minrk:
Got another failure, where the remove strategy was replacing conflicting ADDRANGE vs ADDRANGE, when it should be replacing

Looks like the only (tested) failure remaining is handling the both-sides-add case:

```python
    b = [1, 9]
    l = [1, 2, 7, 9]
    r = [1, 3, 7, 9]
```

where we want

```python
[1, 7, 9]
```

but get

```python
[1, 9]
```

(i.e. the matching ADD 7 is treated as part of the 2 vs 3 conflict). I can definitely imagine cases where this would be annoying, but I don't think it's critical.

martinal:
Fixed all remaining tests and rebased on master.

minrk:
:shipit:

PR #203, opened 2016-12-12
Fix typescript lock to 2.0.x

vidartf:
Missing change to #200, should fix #201.

PR #204, opened 2016-12-12
Appveyor CI

vidartf:
Adds a configuration for running CI on appveyor.

Includes #203 as it would otherwise be prevented from running tests.

TODO:
 - [x] Enable project on appveyor (which account should it run on? This affects who can clear/rebuild etc.)
 - [x] What tests should be included? Currently, it `pip install`s (builds js deps), and runs python tests on python 3.5 (x64) with node 6.9. I'm in favor of keeping it simple, as appveyor does not support parallell jobs on free accounts.
 - [x] Fix any test errors revealed (tests on my appveyor account reveal one error with git merge driver).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,11 @@
 # Do not build feature branch with open Pull Requests
 skip_branch_with_pr: true
 
+# scripts that are called at very beginning, before repo cloning
+init:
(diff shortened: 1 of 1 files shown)


martinal:
This is to get windows coverage? I guess we cover python version differences on travis. There might be windows specific unicode issues with 2.7.

vidartf:
Added an entry for Python 2.7 (32-bit).

vidartf:
All green!

PR #205, opened 2016-12-15
Fix/improve output comparison in diff

vidartf:
Fixes a bug in output comparison (ada5a60), and also makes some changes to speed up the comparison (other commits).


martinal:
Looks good.

PR #206, opened 2016-12-15
Fix JS testing instructions

saulshanabrook:
You have to run JS tests from `nbdime-web` directory, not the root directory.

### DIFF
--- a/docs/source/testing.rst
+++ b/docs/source/testing.rst
@@ -30,6 +30,7 @@ options.
 
 To run javascript/typescript tests, enter::
 
+    cd nbdime-web/
     npm test
 
 from the ``nbdime-web`` folder.


vidartf:
That is the line underneath :)

docs/source/testing.rst
```diff
@@ -30,6 +30,7 @@ options.
 
 To run javascript/typescript tests, enter::
 
+    cd nbdime-web/
     npm test
 
 from the ``nbdime-web`` folder.
```
vidartf: If adding `cd` statement, remove this line.

vidartf: (for clarity, the `from the ``nbdime-web`` folder.` line)

saulshanabrook: Oh good catch! I must have been half asleep when I read this...

PR #211, opened 2016-12-16
[WIP] Fix/improve manual edit

vidartf:
This PR should add proper support for manual editing in the merge pane of the merge view.

This isn't completely ready for merge yet, but it is good enough to share, and to recruit help in breaking it in interesting ways. Indicators for breaking: edited text becomes incorrect; an error notification pops up complaining about model/editor out of sync. Alignment issues are not critical for the end saved result, but might indicate that the decisions/diffs are a little wonky (and might therefore be easy to break with subsequent edits).

Currently this is only supported for a four-way merge view, so I also added a "force four way view" button to the header of all cells that are not four-way. In the future this might be solved more elegantly, but this gets the job done.


vidartf:
Note: This is intended for merge after the next release, as it likely needs more polishing than we are able to perform in time.

Issue #99, opened 2016-10-19
Support attachment field


vidartf:
Version 4.1 of the notebook format added the `attachments` field to raw and markdown cells. Nbdime should also support this field.

### TODO:
 - [ ] Add server side predicates, differs and merge strategy for `attachments` to be similar to that of output mime bundles.
 - [ ] Add display capabilities in diff web view.
 - [ ] Add display capabilities in merge web view.

vidartf:
First point fixed in #139.

PR #215, opened 2016-12-19
Various web improvements

vidartf:
- Clean up paths for static resources.
 - Have diff and merge view start rendering cell widgets as they become ready. Previously, the rendering could be severely delayed from laying out the CodeMirror editors for large notebooks. Some other optimizations were also applied.
 - Improve stylings, particularly of outputs in diff.
 - Refactors renderable models / widgets in preparation for #99.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,4 +21,4 @@ nbdime-web/lib/
 nbdime-web/coverage/
 nbdime-web/test/build/
 
-nbdime/webapp/static/build/
+nbdime/webapp/static/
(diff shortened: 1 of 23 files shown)


vidartf:
Not exactly sure what the CI failure is, as I am not able to reproduce it locally. However, it persisted over a re-run, so we should probably figure it out before merging.

vidartf:
This PR should pass once rebased onto #216. Waiting for that.

vidartf:
Rebased. Had to clear the Travis caches again, but now all green! With #216 merged, I do not think we will have cache issue again for a while (I think the cache might need clearing out after removing dependencies?).

minrk:
If the cache is working properly, it shouldn't ever need clearing. How much time does the cache save?

vidartf:
We cache `node_modules`, which should save quite a bit for the `npm install` step (based on local experience). From Travis timings, it looks like it is >4 min without cache, and >2 min with cache. In principle, the `node_modules` cache should be invalidated when `package.json` changes. I know you can do this with appveyor (but its caches didn't cause any problem now), but have not been able to find the same for Travis.

PR #216, opened 2016-12-20
Update web dependendencies

vidartf:
- Bump the value of various dependencies, particularly jupyterlab/phosphor and adjoining packages. Fix issues, and remove previous workarounds.
 - Switch to using `karma-typescript` for running tests. This simplifies dependencies and coverage generation (reports should also be more accurate).
 - Fix some issues uncovered while updating.
 - Update the version of npm packages to 0.2.0


vidartf:
All green!

Issue #219, opened 2016-12-23
Conflicting default location for git attributes file


baldwint:
When enabling the diff and merge drivers in git with no configured value for `core.attributesFile`, nbdime will write attributes information to a new file at `~/.gitattributes`, even though this is not the default location for the attributes file. The [git documentation](https://git-scm.com/docs/gitattributes) states

> Attributes that should affect all repositories for a single user should be placed in a file specified by the core.attributesFile configuration option (see git-config[1]). Its default value is $XDG_CONFIG_HOME/git/attributes. If $XDG_CONFIG_HOME is either not set or empty, $HOME/.config/git/attributes is used instead.

As a result, a novice user following the [git integration quickstart](http://nbdime.readthedocs.io/en/latest/index.html#git-integration-quickstart) will find that it doesn't work (assuming that they, like me, do not have `core.attributesFile` explicitly set).

baldwint:
Here is a patch that changes this behavior to write to the default location. I wasn't sure how to approach writing an automated test for the `global_` branch of this function, but I manually checked that it fixes the quickstart on my machine (MacOS 10.11.6 El Cap). I have not tested that it works on an OS which uses `$XDG_CONFIG_HOME`.

minrk:
Fixed by #220

PR #220, opened 2016-12-23
Fix default location of global gitattributes file

baldwint:
Git's default location for the attributes file is not `~/.gitattributes`. Update nbdime's default to reflect the one described [here](https://git-scm.com/docs/gitattributes). Addresses #219.

### DIFF
--- a/nbdime/utils.py
+++ b/nbdime/utils.py
@@ -145,7 +145,7 @@ def locate_gitattributes(global_=False):
             if os.environ.get('XDG_CONFIG_HOME'):
                 gitattributes = os.path.expandvars('$XDG_CONFIG_HOME/git/attributes')
             else:
-                gitattributes = os.path.expandvars('$HOME/.config/git/attributes')
+                gitattributes = os.path.expanduser('~/.config/git/attributes')
     else:
         # find .gitattributes in current dir
         path = os.path.abspath('.')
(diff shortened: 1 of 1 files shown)


nbdime/utils.py
```diff
@@ -142,7 +142,10 @@ def locate_gitattributes(global_=False):
             bpath = check_output(['git', 'config', '--global', 'core.attributesfile'])
             gitattributes = os.path.expanduser(bpath.decode('utf8', 'replace').strip())
         except CalledProcessError:
-            gitattributes = os.path.expanduser('~/.gitattributes')
+            if os.environ.get('XDG_CONFIG_HOME'):
+                gitattributes = os.path.expandvars('$XDG_CONFIG_HOME/git/attributes')
```
vidartf: On windows:
```python
>>> print(os.path.expandvars('$HOME/.config/git/attributes'))
$HOME/.config/git/attributes
```

So, I would use `os.path.expanduser(...)` here.

vidartf: PS: I made the comment on the wrong line. This should be for the statement below.

vidartf:
Thanks for spotting this! The fix looks good to me, except for the comment made above.

baldwint:
Thanks for testing this on windows! I have changed back to `expanduser`.

PR #232, opened 2017-01-09
Add `--system` flag for git config commands

### LINKED ISSUE #225: Consider supporting --system flag in tool and driver config
By analogy with `git config --system` it would edit `/etc/gitconfig`.

vidartf:
Resolves #225.

Adds `--system` flag to `config` subcommand of gitdiffdriver, gitdifftool, gitmergedriver and gitmergetool. 

It extracts the shared parts of the `config` subcommand to `args.py` for ease of maintenance, although this leads to somewhat large function calls.

### DIFF
--- a/nbdime/gitdiffdriver.py
+++ b/nbdime/gitdiffdriver.py
@@ -21,21 +21,23 @@ from __future__ import print_function
 import io
 import os
 import sys
-from subprocess import check_call, check_output, CalledProcessError
+from subprocess import check_call, CalledProcessError
 
 from . import nbdiffapp
+from .args import add_git_config_subcommand
 from .utils import locate_gitattributes, ensure_dir_exists
 
-def enable(global_=False):
+
+def enable(scope=None):
     """Enable nbdime git diff driver"""
     cmd = ['git', 'config']
-    if global_:
-        cmd.append('--global')
+    if scope:
+        cmd.append('--%s' % scope)
 
     check_call(cmd + ['diff.jupyternotebook.command', 'git-nbdiffdriver diff'])
-    gitattributes = locate_gitattributes(global_)
+    gitattributes = locate_gitattributes(scope)
     if gitattributes is None:
-        print("No .git directory in %s, skipping git attributes" % path, file=sys.stderr)
+        print("No .git directory in %s, skipping git attributes" % gitattributes, file=sys.stderr)
         return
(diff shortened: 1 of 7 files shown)


nbdime/gitdiffdriver.py
```diff
@@ -21,21 +21,23 @@
...
-        cmd.append('--global')
+    if scope:
+        cmd.append('--%s' % scope)
 
     check_call(cmd + ['diff.jupyternotebook.command', 'git-nbdiffdriver diff'])
-    gitattributes = locate_gitattributes(global_)
+    gitattributes = locate_gitattributes(scope)
     if gitattributes is None:
-        print("No .git directory in %s, skipping git attributes" % path, file=sys.stderr)
+        print("No .git directory in %s, skipping git attributes" % gitattributes, file=sys.stderr)
```
minrk: This message won't make sense. The arg should be the CWD, and the message should only appear for scope=None.

nbdime/utils.py
```diff
@@ -157,6 +182,21 @@ def locate_gitattributes(global_=False):
                 gitattributes = os.path.expandvars('$XDG_CONFIG_HOME/git/attributes')
             else:
                 gitattributes = os.path.expanduser('~/.config/git/attributes')
+    elif scope == 'system':
+        # git docs: "Attributes for all users on a system should be placed in
+        # the $(prefix)/etc/gitattributes file". Our job is then to check for
+        # $(prefix) value.
+        try:
+            with set_env(GIT_EDITOR='echo'):
+                bpath = check_output(['git', 'config', '--system', '-e'])
```
minrk: You can pass an `env=` argument to check_output. You don't need to modify the current process's env to do this.

nbdime/utils.py
```diff
@@ -157,6 +182,21 @@ def locate_gitattributes(global_=False):
...
+                gitconfig = bpath.decode('utf8', 'replace').strip()
+                gitattributes = os.path.join(os.path.dirname(gitconfig), 'gitattributes')
+        except CalledProcessError:
+            # Default to most likely case of empty $(prefix)
+            # Sanity check:
+            if not os.path.exists('/etc'):
+                raise EnvironmentError('Could not find system gitattribute location!')
```
minrk: gitattribute[s]

PR #237, opened 2017-01-12
[WIP] Optimize approx text comparison

vidartf:
Previously, base64 strings embedded in e.g. HTML outputs could blow up the diff time (>1 min, compared to < 10 sec after this PR). To side-step this issue, the `autojunk` feature of `difflib.SequenceMatcher` is enabled.

@martinal Is there a good rational for the autojunk feature being turned off previously? Or was this just a cautionary choice? I'm not entirely sure of the side effects of this, but I'm only enabling it for approximate comparison (not diffing), and it didn't break any test :)

I've also added some of the profiling helpers I used to diagnose this, with a helper text.

### DIFF
--- a/nbdime/diffing/generic.py
+++ b/nbdime/diffing/generic.py
@@ -33,7 +33,7 @@ def default_differs():
     return defaultdict(lambda: diff)
 
 
-def compare_strings_approximate(x, y, threshold=0.7, autojunk=False):
+def compare_strings_approximate(x, y, threshold=0.7, quick=False):
     "Compare to strings with approximate heuristics."
     # TODO: Add configuration framework
     # TODO: Tune threshold with realistic sources
@@ -63,14 +63,19 @@ def compare_strings_approximate(x, y, threshold=0.7, autojunk=False):
 
     # So the heavy ratio function is only used for close calls.
     # s = difflib.SequenceMatcher(lambda c: c in (" ", "\t"), x, y, autojunk=False)
-    s = difflib.SequenceMatcher(None, x, y, autojunk=autojunk)
+    s = difflib.SequenceMatcher(None, x, y, autojunk=False)
 
     # Use only the fast ratio approximations first
     if s.real_quick_ratio() < threshold:
         return False
     if s.quick_ratio() < threshold:
         return False
-    return s.ratio() > threshold
+
+    # Skip slower and stricter check unless if quick is set
+    if quick:
+        return True
+    else:
+        return s.ratio() > threshold
 
 
 def diff(a, b, path="", predicates=None, differs=None):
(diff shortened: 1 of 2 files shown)


martinal:
The autojunk behaviour has some nasty side effects, I turned it off again but tried dropping the most expensive ratio computation for the approximate alignment instead. Seems to work. I also did some other minor fixes/improvements in text alignment. And found a few unrelated bugs that I just fixed in here to save time.

nbdime/diffing/notebooks.py
```diff
@@ -45,73 +46,77 @@
...
-    shortlen = 10  # TODO: Add this to configuration framework
-    if nx < shortlen and ny < shortlen:
-        return True
 
-    return compare_strings_approximate(x, y, threshold=0.7)
+def compare_text(x, y, strict):
+    if strict:
+        return compare_strings_approximate(x, y, threshold=0.95)
+    else:
+        return compare_strings_approximate(x, y, threshold=0.7, quick=False)
```
vidartf: Quick already defaults to False, was this supposed to be True?

vidartf:
Currently, the alignments in the diff are off with this branch (some cells that are quite clearly unrelated get aligned). Not sure what the exact cause it.

nbdime/diffing/generic.py
```diff
@@ -55,14 +60,22 @@ def compare_strings_approximate(x, y, threshold=0.7):
...
+    # Use only the fast ratio approximations first
     if s.real_quick_ratio() < threshold:
         return False
     if s.quick_ratio() < threshold:
         return False
-    return s.ratio() > threshold
+
+    # Skip slower and stricter check unless if quick is set
+    if quick:
+        return True
```
vidartf: Should this maybe return False?

vidartf: Probably not, if I do, very few things get aligned (but there seems to be no false positives).

martinal: Should  be True. But I guess this was a bad idea. The quick ratios are too inaccurate to trust.

martinal: We could try something like this instead:
https://pypi.python.org/pypi/editdistance
but it's a compiled extension.

martinal: I tested this library and for the notebooks you sent me it took more than 10 GB memory before I killed it. Not sure what to do now...

vidartf: I think the old behavior is probably the best, but that (certain text-type) MIME data comparisons can be optimized by splitting on lines.

vidartf: The editdistance package could possibly be an optional dependency (i.e. use it if available)?

vidartf: Haha, 10GB. Let us drop that for now then.

martinal:
Maybe we should just set a size limit on comparing outputs. If len(value) > limit only full equality will be considered equal.

vidartf:
Superseded by #400.

PR #249, opened 2017-01-25
Remove appveyor install of explicit git version

vidartf:
Git should be updated in appveyor/ci#1214, so no longer need to manually perform installation step.


vidartf:
Included fix for Travis, which makes this all green!

PR #255, opened 2017-02-01
Add `--ip` option to web entries

### LINKED ISSUE #244: building the webapp to expose the preliminary api as a service
@vidartf 
I may be jumping the gun here, but is there a basic dockerfile available that will install and then run the nbdime [webapp](https://github.com/jupyter/nbdime/blob/master/nbdime/webapp/nbdimeserver.py).

I manually installed a dev version, following [this](https://nbdime.readthedocs.io/en/latest/installing.html#installing-latest-development-version) and then tried running `python nbdime/webapp/nbdimeserver.py -p 8888`, but wasn't able to get anything running.  Maybe this is still a WIP.

I was experimenting around getting a number of 'services' running under jupyterhub, using kubernetes and wanted to add nbdime as a service in order to potentially be able to access its [API](https://nbdime.readthedocs.io/en/latest/restapi.html).  Is this viable to do, as yet?

Thanks.

vidartf:
Adds an argument `--ip` to all web entry points (and similarly, `nbdimeserver` accepts an `ip` keyword), such that the listening interface can be specified.

Resolves #244.

Security note: As nbdime has no authentication built in, this might lead to exposing parts of your file system to an outsider, so this option should be used with care.

### DIFF
--- a/nbdime/args.py
+++ b/nbdime/args.py
@@ -79,6 +79,10 @@ def add_web_args(parser, default_port=8888):
         default=None,
         type=str,
         help="specify the browser to use, to override the system default.")
+    parser.add_argument(
+        '--ip',
+        default='127.0.0.1',
+        help="specify the interface to listen to for the web server")
     cwd = os.path.abspath(os.path.curdir)
     parser.add_argument(
         '-w', '--workdirectory',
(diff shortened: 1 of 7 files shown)


Analect:
@vidartf 
Tried testing this for my use-case to expose nbdime as a service on jupyerhub, but without success.  See my comments over [here](https://github.com/jupyter/nbdime/issues/244#issuecomment-277925608)

PR #268, opened 2017-02-20
escape encoding errors when presenting diffs/output

### LINKED ISSUE #101: unicode error for notebooks with interactive matplotlib figures
With active diff driver `git diff HEAD~1` fails for notebooks that contain interactive matplotlib windows, i.e., which start with `%matplotlib notebook`

``` sh
$ git diff HEAD~1

Traceback (most recent call last):
  File "/Users/pjuhas/anaconda/envs/dev/bin/git-nbdiffdriver", line 11, in <module>
    load_entry_point('nbdime', 'console_scripts', 'git-nbdiffdriver')()
  File "/Users/pjuhas/programs/github/jupyter/nbdime/nbdime/gitdiffdriver.py", line 117, in main
    show_diff(opts.a, opts.b)
  File "/Users/pjuhas/programs/github/jupyter/nbdime/nbdime/gitdiffdriver.py", line 75, in show_diff
    nbdiffapp.main([before, after])
  File "/Users/pjuhas/programs/github/jupyter/nbdime/nbdime/nbdiffapp.py", line 119, in main
    r = main_diff(args)
  File "/Users/pjuhas/programs/github/jupyter/nbdime/nbdime/nbdiffapp.py", line 44, in main_diff
    pretty_print_notebook_diff(afn, bfn, a, d)
  File "/Users/pjuhas/programs/github/jupyter/nbdime/nbdime/prettyprint.py", line 331, in pretty_print_notebook_diff
    print("\n".join(p))
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2265' in position 1080: ordinal not in range(128)
fatal: external diff died, stopping at test.ipynb
```

The interactive plot window from matplotlib (version 1.5.3) adds the following java code to the notebook:

```
...
       "        alert('Your browser does not have WebSocket support.' +\n",
       "              'Please try Chrome, Safari or Firefox ≥ 6. ' +\n",
...
```

where the `≥` symbol is the \u2265 Unicode character that makes the diff program fail.

minrk:
rather than raising UnicodeEncodeErrors.

closes #101

Need to do a bit of testing on Python 2 and Windows to ensure this is right everywhere.

### DIFF
--- a/nbdime/nbdiffapp.py
+++ b/nbdime/nbdiffapp.py
@@ -80,12 +80,9 @@ def _build_arg_parser():
 def main(args=None):
     if args is None:
         args = sys.argv[1:]
-    if sys.platform.startswith('win'):
-        import colorama
-        colorama.init()
+    setup_std_streams()
     arguments = _build_arg_parser().parse_args(args)
     nbdime.log.init_logging(level=arguments.log_level)
-    setup_std_streams()
     return main_diff(arguments)
 
 
--- a/nbdime/nbdiffapp.py
+++ b/nbdime/nbdiffapp.py
@@ -17,7 +17,7 @@ import nbdime
 from nbdime.diffing.notebooks import diff_notebooks
 from nbdime.prettyprint import pretty_print_notebook_diff
 from nbdime.args import add_generic_args, add_diff_args, add_filename_args
-from nbdime.utils import EXPLICIT_MISSING_FILE, read_notebook
+from nbdime.utils import EXPLICIT_MISSING_FILE, read_notebook, setup_std_streams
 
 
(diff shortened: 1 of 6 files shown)


vidartf:
The CI now reveals some interesting behavior on certain platforms...

Comparing the behavior of this PR without `PYTHONIOENCODING` and with `PYTHONIOENCODING=:backslashreplace` on Windows: They both output the same textual content, but without the env variable, the color codes are printed verbatim to the console...

minrk:
@vidartf ah, that's probably interacting with colorama, which I think also wraps sys.stdout. Running mine first may fix it, or it may need to be more fiddly than that.

PR #275, opened 2017-03-13
add missing #egg=nbdime to pip install -e

### LINKED ISSUE #274: install instructions not working?
I tried
```
pip install -e git+https://github.com/jupyter/nbdime
```
and it yields
> Could not detect requirement name, please specify one with #egg=

I'm not sure -e makes sense with a github repo. Can you explain what the intention is?

minrk:
Avoids "Could not detect requirement name, please specify one with #egg="

closes #274

### DIFF
--- a/docs/source/installing.rst
+++ b/docs/source/installing.rst
@@ -100,7 +100,7 @@ Install the development version
 
 Download and install directly from source::
 
-    pip install -e git+https://github.com/jupyter/nbdime
+    pip install -e git+https://github.com/jupyter/nbdime#egg=nbdime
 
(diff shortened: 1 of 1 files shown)


vidartf:
If I run this here, pip will clone nbdime into `cd/src/nbdime`. The `src` folder will have a file `pip-delete-this-directory.txt` with this content:
```
This file is placed here by pip to indicate the source was put
here by pip.

Once this package is successfully installed this source code will be
deleted (unless you remove this file).
```

I would say this command is acting quite strange...

minrk:
If there's a bug, it's a pip bug that this text file doesn't accurately indicate that the directory should not be deleted if there are editable installs in it. Otherwise, the install is working as intended.

PR #276, opened 2017-03-16
Diff options

### LINKED ISSUE #273: Exclude cell's output from the diff
It would be very nice if there were an option to filter out the cell's output when displaying the diff in the command line.
In this way, one would to see only the changes made in the cell's code or markdown.

vidartf:
Adds flags to diff/merge apps to selectively perform diff/merge on source, outputs, attachments, metadata, or any combination of these. 

Examples:
```
nbdiff a.ipynb b.ipynb -s    # Only considers source for diff

nbdiff a.ipynb b.ipynb -o    # Only considers outputs for diff

nbdiff a.ipynb b.ipynb -O    # Considers everything *except* outputs for diff

nbdiff a.ipynb b.ipynb --ignore-outputs    # Long form of previous

nbdiff a.ipynb b.ipynb -sm    # Only considers source and metadata for diff

nbdiff a.ipynb b.ipynb -sM    # Will give an error, as combination of flags doesn't make sense
```

`nbmerge ` will take the same flags (anything not covered by flags should keep base value).

(Note that due to an detail of how the differ lookup works, cell's `execution_count` does not currently fall into any group, and is therefore always included).

Resolves #273.

### TODO
 - [ ] Document feature

### DIFF
--- a/nbdime/args.py
+++ b/nbdime/args.py
@@ -9,6 +9,8 @@ import os
 
 from ._version import __version__
 from .log import init_logging, set_nbdime_log_level
+from .diffing.notebooks import set_notebook_diff_targets
+
 
 class LogLevelAction(argparse.Action):
     def __call__(self, parser, namespace, values, option_string=None):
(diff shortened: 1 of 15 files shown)


nbdime/args.py
```diff
@@ -17,6 +19,62 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+        elif state != opt and opt is not None:
+            message = 'Arguments must either all be negative or all positive: %r' % (arg_names,)
+            raise argparse.ArgumentError(None, message)
+    if state is None:
+        state = not default
+
+    # Set all unset to the opposite of state (if all specified are True, others are False and vice versa)
+    for name in arg_names:
+        if getattr(ns, name) is None:
+            setattr(ns, name, not state)
```
minrk: It's a little confusing that this is `not not default` in the default case, though I get why. It might be useful to use a more descriptive name for a boolean that came from user args than `state` such as `set_value` or `toggle`.

Maybe it would be a bit clearer above to have:

```python
if toggle is not None:
    # a value was specified, default unspecified items to the inverse
    default = not state
...
    setattr(ns, name, default)

vidartf: Yes, it was kind of hard to follow, cleaned it up.

vidartf:
The tests I added also revealed an issue with outputting diff as JSON on Python 2.7, which I know fixed as well.

PR #279, opened 2017-04-10
get mathjax from cdnjs

### LINKED ISSUE #278: MathJax CDN dependency shutting down April 30
Announcement here: https://www.mathjax.org/cdn-shutting-down/

> Alternative CDN providers
> The easiest way for most site owners will be to simply switch to another CDN provider. We recommend cdnjs which uses CloudFlare for delivery and thus provides the same level of quality as our own CDN. We have been in touch with cdnjs’s maintainers and will help push future MathJax releases to cdnjs.
>
> For example, if you have been using the latest MathJax version (v2.7.0) change
> 

```html
<script type="text/javascript" async
  src="https://cdn.mathjax.org/mathjax/2.7-latest/MathJax.js?...">
</script>
```

> to

```html
<script type="text/javascript" async
  src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.0/MathJax.js?...">
</script>
```

minrk:
cdn.mathjax.org is [shutting down](https://www.mathjax.org/cdn-shutting-down/)

closes #278


vidartf:
That was quite a short warning!

PR #281, opened 2017-04-13
Use diff args for git diff driver

vidartf:


vidartf:
All green with tests :)

PR #282, opened 2017-04-20
Add caching for mimedata comparison

vidartf:
It seems that with the hierarchical algorithm we end up comparing the same things several times. Due to their low level in the hierarchy, and comparatively large size, MIME data comparisons are especially affected by this. This PR addresses this by simply adding an LRU cache on the MIME data comparison.

Related to #237.

### DIFF
--- a/nbdime/diffing/notebooks.py
+++ b/nbdime/diffing/notebooks.py
@@ -3,14 +3,14 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-from __future__ import unicode_literals
-
 """Tools for diffing notebooks.
 
 All diff tools here currently assumes the notebooks have already been
 converted to the same format version, currently v4 at time of writing.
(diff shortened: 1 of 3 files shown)


minrk:
Are there any higher-level comparisons that should be similarly memoized?

vidartf:
I think ideally, you would want the memoization on output comparison, but currently the outputs include lists, which are unhashable. We could ensure that all such lists are tuples before we start processing, but I'm not sure how much other code that would affect.

Another point of improvement could be to make the memoization commutable on the two comparison operands, given that the overhead for such a feature is small (e.g simply sort the two hashes before combining), but I'm not sure the gains are worth the implementation and maintenance effort.

minrk:
Fair enough. Merging this now, but we can explore further memoization later.

Since we shouldn't be modifying notebooks during diff, we could do a one-time transform to immutable / hashable containers (list -> tuple, dict would need a custom hashable subclass, I think).

PR #283, opened 2017-04-20
Refactor setup.py code to use jupyter-packaging

vidartf:
Use setupbase.py from jupyter-packaging.

This should also fix the issue that @martinal experienced where a `pip install .` from the repo would not include the built files (via the implicit call to `update_package_data()`).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,135 +30,22 @@ PY3 = (sys.version_info[0] >= 3)
 import io
 import os
 from glob import glob
-from subprocess import check_call
-
-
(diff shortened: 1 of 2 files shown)


martinal:
Didn't fix the issue though. Vidar is looking into it.

vidartf:
Quite a severe error crept in here, that I fixed in 393bd566c91ff85bf6a042065c013d5a7c79453b...c2ee8eb392369f89fb0b9cdaae458fb6e7b95115

PR #284, opened 2017-04-21
Do not use develop install on CI

vidartf:
Develop install can hide issues with setup scripts, as seen in #283.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,6 @@ before_script:
   # Set up a virtual screen for Firefox browser testing:
   - |
     if [[ $GROUP == js ]]; then
-      cd nbdime-web
       export CHROME_BIN=chromium-browser
       export DISPLAY=:99.0
       sh -e /etc/init.d/xvfb start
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -27,16 +27,12 @@ before_install:
       npm install -g codecov
     fi
 install:
-  - pip install --upgrade -e ".[test]"
-  - |
-    if [[ $GROUP == js ]]; then
-      cd nbdime-web
-      npm install
```
vidartf: Note: The removal of the `npm install` line should also reveal if that step was not completed during `pip install`.

vidartf:
Hmm... The argument for keeping the develop install is that we can use the cache for node modules. How much do we like quick CI? 😅

.travis.yml
vidartf: PS: This should actually only be removed if we keep the develop install, otherwise we need it.

minrk:
Another option is to do a wheel install:

```
python setup.py bdist_wheel
pip install dist/*.whl
```

I don't think the `node_modules` cache is as important as the npm download cache.

vidartf:
Green now :)

PR #287, opened 2017-05-05
Changelog for 0.3

minrk:
and bump mathjax to latest and tweak a few small things in the docs that I noticed.

cc @vidartf @JamiesHQ


vidartf:
Neat. Nice catch on the typos!

PR #290, opened 2017-05-22
Fix mishandled ignore flags

baldwint:
When gitdiffdriver hands off to nbdiffapp, ignore flags (like`--ignore-metadata`) are dropped because of a bug that does not distinguish between False (an ignore flag is passed) and None (neither an ignore flag nor its opposite are passed). This prevented use of ignore flags from being usable in a gitconfig file.


vidartf:
Thanks for spotting this, and for the fix! I pushed a commit to reuse some similar code from elsewhere in the code base (but there was nothing wrong with the submitted fix as such).

PR #297, opened 2017-07-05
Notebook server/frontend extension

vidartf:
This PR:
 - Adds a notebook server extension with:
   - Difftool support
   - Diff API support (needed for diff tool also)
   - Difftool supports diffing file against git HEAD (normally git files has only been processed on entry points).
   - API for querying if a file is in a git repo
 - Adds a notebook front-end extension:
   - Adds a button/action to diff the saved file against checkpoint (not tested with checkpoint manager).
   - Adds a button/action to diff the saved file against git HEAD, if the notebook is in a git repository.

### TODO:
 - [x] Document installation/use
 - [x] Add unit tests for server extension
 - [ ] ~~Add unit tests for notebook front-end extension~~
 - [x] Ensure front-end extension is included in python package


vidartf:
Preview: 

nbext-preview

vidartf:
Future improvement suggestions:
 - [ ] Use header labels similar to that of the JL plugin in #298.
 - [ ] Allow diffing two notebooks from the tree view by selecting them (labels as suggested above will ensure ordering is not significant).

vidartf:
Rebased on top of #310 and #311.

nbdime/__main__.py
```diff
@@ -82,6 +82,11 @@ def main_dispatch(args=None):
             diff_tool(args) or
             merge_tool(args)
         )
+    elif cmd == 'reg-extensions':
+        # Register nbdime extensions
+        call('jupyter serverextension enable --py nbdime'.split() + args)
+        call('jupyter nbextension install --py nbdime'.split() + args)
+        call('jupyter nbextension enable --py nbdime'.split() + args)
```
minrk: `nbdime reg-extensions` results in `local variable main undefined`. Probably missing a `return` here

minrk:
Files are located relative to:

```python
self.contents_manager.root_dir
```

not necessarily the cwd. So that should be taken into account when resolving paths.

minrk:
For diffing with checkpoints: use contents manager API to get content rather than making assumptions about the filesystem. Something like:

```python
model = yield gen.maybe_future(cm.get('path', content=True, type='notebook'))
checkpoints = yield gen.maybe_future(cm.list_checkpoints('path'))
checkpoint_model = yield gen.maybe_future(cm.get_notebook_checkpoint(checkpoints[0], 'path'))
nbdiff(model['content'], checkpoint_model['content'])
```

Also note that `contents_manager.root_dir` cannot be assumed to be defined for non-file-based contents managers. For these, git diffs should gracefully be unavailable, but checkpoints should still work as long as the above is done.

nbdime/__main__.py
vidartf: Fixed.

vidartf:
When I try the checkpoint code above, I get the following errors:
```python
    [...]
    checkpoint_model = yield gen.maybe_future(cm.get_notebook_checkpoint(checkpoints[0], base))
AttributeError: 'LargeFileManager' object has no attribute 'get_notebook_checkpoint'
```
Any ideas?

vidartf:
Changing the line to 
```python
checkpoint_model = yield gen.maybe_future(
            cm.checkpoints.get_notebook_checkpoint(checkpoints[0], base))
```
produces this error instead:
```python
AttributeError: 'FileCheckpoints' object has no attribute 'get_notebook_checkpoint'
```

I think setting `_checkpoints_class_default` in notebook to `GenericFileCheckpoints` should solve this, but I'm not sure if this is something that is OK to PR to notebook?

PR #298, opened 2017-07-11
Add jupyter lab extension

vidartf:
Builds on the server extension in #297, and adds a jupyter lab extension.

### DIFF
--- a/packages/labextension/package.json
+++ b/packages/labextension/package.json
@@ -31,7 +31,18 @@
     "notebook"
   ],
   "jupyterlab": {
-    "extension": true
+    "extension": true,
+    "discovery": {
+      "server": {
+        "base": {
+          "name": "nbdime"
+        },
+        "managers": [
+          "pip",
+          "conda"
+        ]
+      }
+    }
   },
   "scripts": {
     "clean": "rimraf lib && rimraf dist",
--- a/packages/labextension/package.json
+++ b/packages/labextension/package.json
@@ -1,0 +1,58 @@
+{
+  "name": "nbdime-jupyterlab",
+  "version": "0.1.0",
+  "description": "A JupyterLab extension for showing Notebook diffs.",
+  "main": "lib/index.js",
+  "typings": "lib/index.d.ts",
+  "dependencies": {
+    "@jupyterlab/apputils": "^0.16.0",
+    "@jupyterlab/coreutils": "^1.1.2",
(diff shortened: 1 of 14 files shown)


vidartf:
Future improvement suggestions:
 - [ ] Add action to diff two notebooks (dialog with two filebrowsers?)
 - [ ] Use better icons + style them similarly to existing NB toolbar icons.

These should not be in this PR though.

Analect:
@vidartf ... is this in a working state to use against jupyterlab 0.27?
If I have an external service on jupyterhub running nbdime, can it use that, or does it end up installing nbdime as part of the extension installation?  Tks.

vidartf:
@Analect I cannot guarantee anything about this, but hope to work further on it this month. I did have a working version at some point though, so if you play around with it, you might be able to get it working.

Note: Even after getting the classic notebook extension into master, we might still hold off on this until the relevant jupyterlab APIs reach version 1.0. This would be in order to reduce our own maintenance burden in order to stay compatible with latest JL version.

Analect:
Thanks for the update @vidartf 
To what extent would this extension interact with the [jupyterlab-git](https://github.com/jupyterlab/jupyterlab-git/issues/163) one?
I know you worked a little with @zzhangjii at JupyterCon on this, and was wondering how you were thinking about the integration of the two ... if at all.

vidartf:
@Analect The solution that was agreed and implemented there is for jupyterlab-git to expose a function for diff providers to register themselves. Nbdime would then check on load whether jupyterlab-git is present, and if so, register itself.

vidartf:
This will likely stay as a PR until various jupyterlab APIs stabilize (tip 1.0), in order to reduce maintenance burden of following JL releases.

elgalu:
is it stable enough now?

vidartf:
The extension is still dependent on the following jupyterlab extensions that have not reached a 1.0:
```
@jupyterlab/application
@jupyterlab/apputils
@jupyterlab/docregistry
@jupyterlab/codeeditor
@jupyterlab/codemirror
@jupyterlab/notebook
@jupyterlab/outputarea
@jupyterlab/rendermime
```
This means that for every minor release of those packages, we would have to release a new version of the nbdime lab extension. Hopefully that will not happen too often, but at least I want to wait until https://github.com/jupyterlab/jupyterlab/pull/3608 is included in lab to minimize installation troubleshooting.

PR #300, opened 2017-08-16
Crop image to cell height

jbwhit:
The first few times that I read the documentation, I found the text at the top of the `docs/source/images/nbdiff-web.png` image to be confusing/distracting to what was being conveyed. Especially the "Loading Matplotlib demos..." text.

This pull request is a single modification of the image to crop it to just the jupyter cell under consideration that removes the extraneous parts.


minrk:
Makes sense!

PR #310, opened 2017-10-31
Update web dependencies + use lerna

vidartf:
- Updates the js dependencies (including jupyterlab)
 - Minor fixes to js code
 - Refactor web code to use lerna for eased development


vidartf:
Note: This is intended as the ground work for notebook / jupyterlab extensions, so #297  and #298 will be rebased on top of this.

PR #315, opened 2017-11-04
Fix apparent typo of nbdiff-web

einon:
This looks like a typo where 'nbdime-web' was entered, which doesn't exist, instead of 'nbdiff-web'.


vidartf:
Thanks!

PR #318, opened 2017-11-07
Add --color-words to CLI diff entry points

### LINKED ISSUE #293: Diff does not show indentation changes
I am using nbdime and I had the following problem with it today. I was editing this Python code:

```python
with open("numbers.txt") as f:
    l = []
    for line in f:
        l.append(float(line.strip()))
    l.sort()
    l.reverse()
```

To this code:

```python
with open("numbers.txt") as f:
    l = []
    for line in f:
        l.append(float(line.strip()))
l.sort()
l.reverse()
```

Ie I changed the indentation of the last two list. However, the diff looked like this:

```
## modified /cells/2/source:
@@ -13,5 +13,5 @@ with open("numbers.txt") as f:
    l = []
    for line in f:
        l.append(float(line.strip()))
l.sort()
l.reverse()
```

with no coloring at all, which made me think nothing had changed. I think indentation changes should be shown much more prominently, particularly for a Python tool!

vidartf:
Adds an argument to CLI diff entry points, which controls whether any calls out to `git diff` will use the `--color-words` argument. Off by default, due to its white-space stripping nature, c.f. #293, but can now be turned back on manually (opt-in).

Fixes #293.

### DIFF
--- a/nbdime/args.py
+++ b/nbdime/args.py
@@ -223,6 +223,17 @@ def add_diff_args(parser):
 diff_exclusives = ('sources', 'outputs', 'attachments', 'metadata', 'details')
 
 
+def add_diff_cli_args(parser):
+    """Adds a set of arguments for CLI diff commands (i.e. not web).
+    """
+    parser.add_argument(
+        '--color-words',
+        action='store_true', default=False,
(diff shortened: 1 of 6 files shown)


cfbolz:
Great, thanks a lot!

PR #319, opened 2017-11-08
Web: Render fixes

vidartf:
- Updates MIME type check to follow update in jupyterlab dependency.
 - Ensures images in unchanged outputs are centered, *without* centering unchanged text outputs.

### DIFF
--- a/packages/nbdime/src/styles/diff.css
+++ b/packages/nbdime/src/styles/diff.css
@@ -134,10 +134,12 @@
   width: initial;
   max-width: 100%;
 }
-.jp-Notebook-diff .jp-Diff-twoway .jp-Cellrow-outputs .jp-Diff-unchanged {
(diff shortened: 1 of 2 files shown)


packages/nbdime/src/styles/diff.css
```diff
@@ -134,10 +134,12 @@
   width: initial;
   max-width: 100%;
 }
-.jp-Notebook-diff .jp-Diff-twoway .jp-Cellrow-outputs .jp-Diff-unchanged {
+/* Ensure unchanged images are centered */
+.jp-Notebook-diff .jp-Diff-twoway .jp-Cellrow-outputs .jp-Diff-unchanged .jp-Rendered-Image {
```
minrk: `.jp-RenderedImage`

PR #331, opened 2017-11-28
 Add periodic poll on server 

vidartf:
Ensures that server checks for signals every 5 seconds on Windows. Needed to pick up e.g. Ctrl + C events.


vidartf:
Note: I want to double check that we don't get unwanted log output somewhere due to this before merging.

vidartf:
Splitting PR instead to separate concerns.

PR #334, opened 2017-11-30
Ignoring noisy standard metadata on --ignore-details

ziembla:
That would be great to have more control over what can be ignored in comparisons.

Perhaps at least some "standard and obvoiusly noisy" attributes should be dropped from metadata when "--ignore-details" is used?

I propose a static list below, but it'd be great if that could be done through some kind of configuration...

(I am sorry I was unable to pass the pytest suite because my global git configuration was lacking and some seamingly test-purpose repos were not created. Still, I hope I don't break anything important, though literal comparisons may differ if attributes I like removed were present...)

### DIFF
--- a/nbdime/diffing/notebooks.py
+++ b/nbdime/diffing/notebooks.py
@@ -548,8 +548,12 @@ def set_notebook_diff_targets(sources=True, outputs=True, attachments=True, meta
     metadata_keys = ("/cells/*/metadata", "/metadata", "/cells/*/outputs/*/metadata")
     if metadata:
         for key in metadata_keys:
-            if key in notebook_differs:
-                del notebook_differs[key]
+            if details:
+                if key in notebook_differs:
+                    del notebook_differs[key]
+            else:
+                notebook_differs[key] = diff_ignore_keys(
+                    inner_differ=diff, ignore_keys=['collapsed', 'autoscroll', 'deletable', 'editable'])
     else:
         for key in metadata_keys:
             notebook_differs[key] = diff_ignore


ziembla:
As far as I understand the appveyor logs the failure isn't caused by my commit, the problem seems to be with environment setup done by npm (problems around "lerna bootstrap --hoist")?

vidartf:
Thanks for the suggestion, and your contribution! I agree something like this would be useful. I'll look into the appveyor issue separately.

I'm also wondering if a complementary solution would be to allow for accepting generic ignores, and making nbdime configurable. Then you could pass an option (via flag or config file entry) along the lines of `ignore-path = /cells/*/metadata/autoscroll`. It would be more verbose, but also more powerful.

Which would you prefer to have, only one of the two, or both?

ziembla:
Thanks for your reply!

If the "one" is having list of details expanded and "two" is configurability like suggested, I'd love to have configurability but it that's not coming soon, I'd be glad to have something like my humble expansion included for the time being :-)

The fact is that the list of possibly unimportant elements would never be complete and what's more there'd not be one list that suits all... I'm thinking about cases like you get somebody's notebook and he's got some extension-related metadata for instance that you want to ignore not using the extension for example...

I didn't dig too deep in your code but from what I understand that "configurable" step would be some post- (pre-?) processing as it don't quite fit in the nested "predicate/key" model?

I hope I dont digress too far here but in fact before looking into nbdime's source I thought that adding a git filter (clan/smude one) would be a simpler solution but hit the wall. It looks like the filter is being called before diff (forwarded to nbdiff) but it seems that nbdiff is processing the git file stored somewhere in /tmp not the output of the filter. Perhaps you could shed some light here just off the top of your head?

vidartf:
Thanks for the detailed feedback! Using a filter is inherently different than filtering the diff, as it also prevents something from being committed in the first place. The goal would be for the user to have the option of doing both (i.e. filter generically with nbdime, and have nbdime work with filters beforehand). I did not know there was an issue with nbdime not working with filters (have not tested it myself), I will have to look into that! It might be that there is a limitation with git, if so we should try to make a workaround.

What tool did you use for testing git filtering? An existing package, or a custom one?

vidartf:
> I didn't dig too deep in your code but from what I understand that "configurable" step would be some post- (pre-?) processing as it don't quite fit in the nested "predicate/key" model?

The only extra bit of information that is needed is whether the last part of the path is an immutable on a dict, or a dict/list. This would have to be reflected in the input somehow I guess. After that it is just a matter of mapping the paths in `notebook_differs` to the relevant ignore function (`diff_ignore_keys` for immutables, `diff_ignore` for containers). I'll open a separate issue/PR for this to keep this one from going too off-topic.

ziembla:
Thanks.
As to the testing - I've beed experimenting inspired by <https://gist.github.com/pbugnion/ea2797393033b54674af>.

Diffing changes in notebooks form other people was kind of pain for some time, but I now got to it along the lines of preparing a docker environment to run samples from one ML book... so I can point you there.

In the directory in the commit <https://github.com/ziembla/handson-ml/tree/68e17dcc704ca5b6943bb60042b50bc7ccf39150/docker> there is a "metadata dumping and damaging" experimental `ipynb_cleaner.py` which tries to modify the notebook document. The registration is done at the end of `Dockerfile`.

You'll find it easily even if not versed in docker ;-) Or you can clone the repo (it's my working dev-woj branch though), build and run the image, change a notebook and see the diff in action if you want to ;-)

Still, the result is that json is modified by the filter and dumped to stdout but the nbdiff doesn't see the changes. I didn't check the git's sources yet, but I imagine these should be in the pipeline :-D

vidartf:
I've actually encountered this problem before: https://github.com/kynan/nbstripout/issues/44. Citing the relevant part:

> external diff drivers compare the working tree file with a smudged repo file[...]. It doesn't appear to be currently possible to override this in git.

On the other hand, the filter will control what actually gets committed. 

As is pointed out, you can make a custom script that ensures the diff driver gets the filtered version for diffing:

```bash
#!/bin/bash

# pass arguments through to git-nbdiffdriver, but pass 
# a stripped version of the working tree file
git-nbdiffdriver diff $1 $2 $3 $4 <(cat $1 | nbstripout) $6 $7
```

and use that for the diff driver. 

A possible solution here would be for nbdime to interrogate git whether a filter is applied, and apply that internally before processing (possibly putting this behavior behind a flag).

ziembla:
Thank you for all the details, I'd not find them that easily myself! Still, while the workaround with stdin redirected from "stripping pipeline" shows the unix paradigm "you can always put the small blocks in a way that matches your needs" it's kind of hacky... But's good to know, thanks.

Well, I expect that can be argued both ways, but it seems to me that alternating "the thing" stored in git is not what is really best in the scenario with jupyter notebook cleaning.

Which brings us back to the beginning of the conversation - configurable igoring things in comparisons is useful in more use cases than just git commit!

In all cases below I'd be useful to suggest "what you care about seeing":
- diff with git
- diff between files
- diff with checkpointed version

You might have noticed that for the last case I've created a simple funcion in my basrc... perhaps nbdiff could do similar thing given some flag (kind of off-topic again?)

```
nbd() {
	DIRNAME=$(dirname "$1")
	BASENAME=$(basename "$1" .ipynb)

	WORKING_COPY=$DIRNAME/$BASENAME.ipynb
	CHECKPOINT_COPY=$DIRNAME/.ipynb_checkpoints/$BASENAME-checkpoint.ipynb

	nbdiff "$CHECKPOINT_COPY" "$WORKING_COPY"
}
```

vidartf:
This has been superseeded by #355.

Now you can add the following config (`nbdime_config.json` on `jupyter --paths` somewhere):

```json
{
  "GitDiff": {
    "/cells/*/metadata": ["collapsed", "autoscroll", "deletable", "editable"]
  }
}
```

If you could test that this works as you would expect, it would be very helpful to get that feedback before the next release!

vidartf:
Note: The dff driver now also supports git filters after #335, so that can also be used.

PR #335, opened 2017-12-04
Add git filter integration

vidartf:
- Adds capability for applying git filters according to git config.
- Applies filter on remote as it should for calls to `nbdiff` with git refs.
- As diff drivers/tools cannot correctly apply filters, it will always apply any filter on remote if the flag `--use-filter` is given, otherwise it will not consider filters.

### DIFF
--- a/nbdime/args.py
+++ b/nbdime/args.py
@@ -232,13 +232,42 @@ def add_diff_cli_args(parser):
     )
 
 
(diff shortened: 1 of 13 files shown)


vidartf:
This could still benefit from tests on `nbdiff` with gitrefs, and on the git difftool.

PR #260, opened 2017-02-09
support base-url

minrk:
Resolves #244 .

TODO:

- [x] make sure base_url is supported in every web application
- [x] prefix URLs with base_url in typescript
- [x] run tests with base URL, e.g. `a%20b` to ensure double-escapes don't happen

PR #342, opened 2018-01-18
Making nbdime work with JupyterHub

sat28:
I have successfully updated it and works fine in Jupyter as well as for JupyterHub. 

The issue occurs as the way base_url is formed in JupyterHub is different from Jupyter. Error in console was that /api/isgit not found.


minrk:
Thanks for the PR! I don't think this patch can have fixed anything. Did you perhaps update nbdime while working on it? Base url handling was fixed in #260. If 'base_url' is defined, `get` and `[]` return the same thing. If it's not defined, this PR will error and the `get` will return`/`, which is what we want.

PR #354, opened 2018-03-22
Add config system

vidartf:
Add a config system. Currently it only adds the appropriate CLI arguments.

This differs from the config system used elsewhere in Jupyter by: Adding different configurable sections based on entry points, and entry point categories.

### DIFF
--- a/nbdime/config.py
+++ b/nbdime/config.py
@@ -5,19 +5,21 @@ from ipython_genutils import py3compat
 from jupyter_core.paths import jupyter_config_path
 
 from traitlets import Unicode, Enum, Integer, Bool, HasTraits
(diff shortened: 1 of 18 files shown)


nbdime/config.py
```diff
@@ -0,0 +1,290 @@
...
+_config_cache = {}
+def config_instance(cls):
+    if cls in _config_cache:
+        return _config_cache[cls]
+    instance = _config_cache[cls] = cls()
+    return instance
+
+
+def _load_config_files(basefilename, path=None):
+        """Load config files (py,json) by filename and path.
```
minrk: leftover comment - only json files are loaded

PR #355, opened 2018-03-22
Custom ignore config

vidartf:
Allows for using the config system for full configuration of keys to ignore. This should be a better long term solution than adding more logic to ignore flags on the CLI.


vidartf:
Builds on #354.

PR #356, opened 2018-03-22
Don't use get on defaultdict

vidartf:
The get method side-steps the default factory on the defaultdict, as specified in python docs. This prevented the use of `diff_sequence_multilevel` in certain cases, and of `diff_attachments` on attachments.


vidartf:
Actually, `diff_sequence_multilevel` was never prevented, but it was called inefficiently (it went via `diff`, `diff_list` and its check for predicates).

PR #400, opened 2018-06-20
Optimize text comparison for text MIME data

vidartf:
Supersedes #237.

Difflib's ratio() method blows up for very large contents. This PR limits this problem for output comparison by giving a maximum length for text data when doing mimedata comparison. If both sides of the comparison are larger than this length, it will revert to strict equality instead of the slow `ratio()` method.


vidartf:
TODO: Autogenerate a notebook pair with big text data output to compare, and write some regression tests for it.

PR #457, opened 2019-02-28
set allow_remote_access when listening on non-localhost ip

### LINKED ISSUE #455: 403: Forbidden Error
Hello!

I'd like to reach the nbdiff-web interface over my LAN. Therefore, I tried to start it with:
`nbdiff-web --ip 192.168.0.13 -p 9999`

However, when I try to access it over LAN, I get the following error:

> Blocking request with non-local 'Host' 192.168.0.13 (192.168.0.13:9999). If the notebook should be accessible at that name, set NotebookApp.allow_remote_access to disable the check.

I already tried to set `c.NotebookApp.allow_remote_access=True` in the jupyter_notebook_config.py, but that did not work (which did not really suprise me). However, I dont know where else I should put it, because there does not seem to be an equivalent file for nbdime. So how do I have to set up nbdime, in order to reach it over LAN?

Kind regards and thanks in advance!

minrk:
otherwise the notebook handlers block access as a CORS / DNS-rebinding protection

closes #455

### DIFF
--- a/nbdime/webapp/nbdimeserver.py
+++ b/nbdime/webapp/nbdimeserver.py
@@ -336,8 +336,8 @@ def make_app(**params):
         'base_url': base_url,
         'jinja2_env': env,
         'mathjax_url': prefix + '/nb-static/mathjax/MathJax.js',
-        'local_hostnames': ['localhost', '127.0.0.1']
-        }
+        'local_hostnames': ['localhost', '127.0.0.1'],
+    }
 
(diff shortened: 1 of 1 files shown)


maxfrei750:
@minrk @vidartf Thanks for fixing the problem so quickly. Is there an ETA for the next release, where this fix would be incorporated?

vidartf:
Hopefully this week.

vidartf:
Released.

PR #463, opened 2019-03-01
change the order of if statements so that missing is checked first

Godley:
Found running this in a container in linux, if `/dev/null` exists on the os then it will try to read from it and tip over.


vidartf:
Thanks!

Issue #456, opened 2019-02-26
add syntax highlighting to code cells in nbshow output 


lebedov:
It would be nice to add syntax highlighting to code cells in nbshow output using something like pygments (proof-of-concept available [here](https://gist.github.com/lebedov/bf24781b891de609f1a930132941211e)).

vidartf:
Hi! That could be a very nice feature, indeed. While I have a bit of hesitation to put too many features into a "side-tool" of nbdime, it does look nice and compact. Do you want to open a PR? The best implementation would probably be a function call before [this call](https://github.com/jupyter/nbdime/blob/435ce3b2d96f8194ea60448b4de00a47a394a7c6/nbdime/prettyprint.py#L616). It should ideally take an argument specifying the language, so that the correct lexer can be chosen (I can help in threading that argument correctly through the call chain).

lebedov:
PR created. Please update to pass the source language through the call chain so as to select the correct lexer (which is currently hard coded to Python).

vidartf:
Closed by #466.

PR #466, opened 2019-03-11
Colorize source with pygments before displaying.

lebedov:
Implements feature requested in https://github.com/jupyter/nbdime/issues/456. Currently hard coded to colorize source as Python; needs to be parameterized to pass language to `find_lexer_class_by_name()` so as to select the appropriate lexer.

### DIFF
--- a/nbdime/prettyprint.py
+++ b/nbdime/prettyprint.py
@@ -25,9 +25,6 @@ except ImportError:
     from backports.shutil_which import which
 
 import colorama
-from pygments import highlight
-from pygments.formatters import Terminal256Formatter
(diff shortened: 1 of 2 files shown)


vidartf:
Thanks! I added a few points:
- Made dependency on pygments optional.
- Only highlight if we have no prefix (e.g. for diffing etc.) as the highlighting would interfere with the current color use.
- Added logic for getting pygments lexer from notebook metadata. It falls back on language name if not present.
- Uses "markdown" for markdown cells.
- Handles missing lexer class exceptions.

lebedov:
Looks good.

PR #470, opened 2019-04-17
Upgrade to jupyterlab 1.0.0a3

gnestor:
Upgrade jupyterlab dependencies to work with 1.0.0a3


vidartf:
Thanks for this. Would you mind removing the added lock files?

gnestor:
Ok updated!

vidartf:
Any idea what is going on in the package-lock.json file? It seems like a change in dedupe logic. Which version of npm are you using?

gnestor:
Not sure. I'm running npm 6.9.0.

vidartf:
superseeded by https://github.com/jupyter/nbdime/pull/483

PR #475, opened 2019-05-16
Fix styling issue

vidartf:
Fix a regression after #462.


vidartf:
The issue was mainly that the codemirror panes did not get the correct `width` styling, causing overflow with line-wrapping.

PR #477, opened 2019-05-20
Use config for webapps

### LINKED ISSUE #476: Config port being ignored
I've created a file `nbdime_config.json`:

```json
{
	"NbDiffWeb": {
		"port": 40135
	},
	"NbMergeWeb": {
		"port": 40135
	},
	"Server": {
		"port": 40135
	}
}
```

I verify it's being used by 
```bash
> nbdiff-web --config
NbDiffWeb:
  Ignore: {}
  attachments: <unset, resolves to true>
  base_url: "/"
  browser: null
  color_words: false
  details: <unset, resolves to true>
  ip: "127.0.0.1"
  metadata: <unset, resolves to true>
  outputs: <unset, resolves to true>
  persist: false
  port: 40135
  sources: <unset, resolves to true>
  workdirectory: "/home/tyler/Notebooks"
```

However, port 40135 is not used:

```bash
> nbdiff-web notebookA.ipynb notebookB.ipynb
[I nbdimeserver:377] Listening on 127.0.0.1, port 44087
```

CMD flag works though:
```
> nbdiff-web --port 40135 notebookA.ipynb notebookB.ipynb
[I nbdimeserver:371] Listening on 127.0.0.1, port 40135
```

vidartf:
Fixes #476 .

Not obvious to me why these were not included in the first place. Did we forget about them, or did we leave them out for some reason? If so, we have not been consistent in excluding them.

@minrk Do you remember anything about this? I have an echo of a memory that we might have done this on purpose for some reason, but I cannot for the life of me remember what that reason could be, and I find no clues in code / git history either.

### DIFF
--- a/nbdime/webapp/nbdifftool.py
+++ b/nbdime/webapp/nbdifftool.py
@@ -5,9 +5,9 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import sys
-from argparse import ArgumentParser
 
 from ..args import (
+    ConfigBackedParser,
     add_generic_args, add_diff_args, add_web_args, add_filename_args,
     args_for_server, args_for_browse, process_diff_flags)
 from .nbdimeserver import main_server as run_server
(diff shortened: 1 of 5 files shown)


vidartf:
Xref #354, #390.

PR #481, opened 2019-06-13
Some UI improvements in Diff widget

jaipreet-s:
As discussed in #469 , this change implements few of the UX improvements

* Coloring changes for added/removed/outputs
* Remove borders around the added/removed lines
* Add border around the "add-remove" panel to contain the whitespace that occurs when a certain cell has been completely removed or added.
* UI changes for "Hide unchanged cell(s)" and the "Outputs changed/unchanged" collapsible panel.

cc @vidartf @weihwang 

-----
Screen Shot 2019-06-13 at 1 56 45 PM


vidartf:
Thanks for this! One issue I see is that the new color/alpha combinations doesn't work very well with the dark theme in jupyterlab:

image

See also [my comment](https://github.com/jupyter/nbdime/issues/469#issuecomment-490905690) on the issue.

jaipreet-s:
Good find. @weihwang - WDYT?

packages/nbdime/src/styles/variables.css
```diff
@@ -2,17 +2,17 @@
...
 
   --jp-diff-added-color0: rgba(0, 200, 20, 0.6);
   --jp-diff-added-color1: rgba(0, 200, 20, 0.3);
-  --jp-diff-added-color2: rgba(0, 200, 20, 0.3);
+  --jp-diff-added-color2: rgba(225, 244, 221, 0.3);
 
   --jp-diff-deleted-color0: rgba(240, 20, 0, 0.55);
   --jp-diff-deleted-color1: rgba(255, 20, 0, 0.3);
-  --jp-diff-deleted-color2: rgba(255, 20, 0, 0.3);
+  --jp-diff-deleted-color2: rgba(244, 221, 221, 1.0);
```
vidartf: Actually, I think this is the only color that looks bad in dark theme because of the increased alpha.

weihwang: I can come up with another set of colors for dark UI

vidartf: This is technically possible due to lab tagging the DOM with dark/light info, but note that this should work for a wide range of syntax highlighting colors as well.

vidartf: Also, the dark/light tag is more of a last resort kind of mechanism. So if you find a color specification that works on both dark and light, that would be best I think.

weihwang: image
Revised color palette to work for both dark/light

vidartf: 🔥

jaipreet-s: Addressed in latest commit

jaipreet-s:
HI @vidartf ,

Just checking in if there's anything you need me to update here or are we good to merge. Thanks!

vidartf:
@jaipreet-s I've been on vacation, so apologies for the slow response-time :) I'm working through this locally as there were some issues with the CSS (the end result did not match the colors posted by @weihwang). It is also related to the 1.0 lab update, as that has some CSS changes.

vidartf:
@weihwang I realized all you mockups are missing a useful feature: Highlighting of the specific words within a line that has changed. E.g. in my screenshot above, it is highlighting the ` + 550000` part. My initial though here is to ensure that the highlighted words will match your colorspec, and have the line-highlight for non-changed parts be significantly dimmed in comparison. Thoughts?

jaipreet-s:
Closing as this is > 2 years old.

Issue #480, opened 2019-06-11
NBDime REST API changes for git diffs


jaipreet-s:
Hi @vidartf and others,

I'm touching base on the work proposed to add visual diff in the Git extension in https://github.com/jupyterlab/jupyterlab-git/issues/331

### Current world

The nbdime REST API  (`/nbdime/api/diff`) provides diff for at a file level

1. Two versions of a notebook file (i.e. specifying base and remote)
2. passing in base as `git:$FILE_NAME` and that computes the diff between version of the file in the Git *working directory* and *HEAD*

### Proposed changes

To support diffs across the Git workflow, the proposal was to enhance this API to allow diff'ing between multiple points such as between two arbitrary git refs(equivalant of `git diff $OLD_COMMIT $NEW_COMMIT` or Diffing the changes in the staging area (equivalent of `git diff --staged`)

There are two ways to go about this

1. Add fields to the existing diff API that allow this functionality
    * Pros: Few changes needed to nbdime
    * Cons: Since the current nbdime API works on a single file, the Git extension has to find all the files that changed between the two git refs and invoke the nbdime API multiple times

2. Add a specific APIs for Git diffs that provide diff for all files based on the Git context passed
    * Pros: 1 API call from the client perspective and no need to pass the file name
    * Cons: Relatively more changes needed to nbdime and  a bit specific to Git

Both will work but I wanted to gather any feedback before embarking on this work.

vidartf:
Note: For alternative 1. there is some work already in a branch on my repo (https://github.com/jupyter/nbdime/compare/master...vidartf:jlgit). Although it would probably be better to not use the prefix method, if it can be avoided.

For alternative 2, I'd appreciate thoughts on pros/cons of putting this in nbdime, vs implementing this somewhere else (e.g. jupyterlab-git).

jaipreet-s:
> Note: For alternative 1. there is some work already in a branch on my repo (master...vidartf:jlgit). Although it would probably be better to not use the prefix method, if it can be avoided.

I agree - If we go along this route, we'll update the API to have explicit fields for git refs

> For alternative 2, I'd appreciate thoughts on pros/cons of putting this in nbdime, vs implementing this somewhere else (e.g. jupyterlab-git).

Good point. One consideration on option 2 is where it is implemented.

**Pro** of implementing this in the Git extension:  

* The Git logic is encapsulated in the Git extension
* We'd likely move faster from the Git extension's perspective and have to worry less about breaking nbdime users

[...]

PR #482, opened 2019-06-18
Add a `/gitdiff` API 

jaipreet-s:
#### Overview
This change adds a server endpoint to diff two git refs for a single file. This is flexible to perform diffs across various points in the Git lifecycle such as diff'ing changes in the Working Tree, Index, or two arbitrary commits. There are two reserved refs WORKING and INDEX to designate the Working Tree and the git Index

A separate API was added to avoid using the "git:" prefix to designate git diffs and add many optional parameters in the existing `/api/diff`. This is backwards compatible, the existing `/api/diff` will work as expected.

Addresses #480 

```
curl -X POST \
  http://localhost:8888/nbdime/api/diff \
  -H 'Content-Type: application/json' \
  -H 'cache-control: no-cache' \
  -d '{
    "base": "git:Untitled1.ipynb"
}'
```

#### Other details

* Submitting early to get feedback. I'll add tests if the overall approach looks fine

#### Testing scenarios
Various cases for the `git diff` API

(Diff all changes in Working Tree)
```
curl -X POST \
  http://localhost:8888/nbdime/api/gitdiff \
  -H 'Content-Type: application/json' \
  -H 'cache-control: no-cache' \
  -d '{
	"ref_prev": "HEAD",
	"ref_curr": "WORKING",
	"file_name": "Untitled1.ipynb"
}'
```

(Diff all changes in staging)
```
curl -X POST \
  http://localhost:8888/nbdime/api/gitdiff \
  -H 'Content-Type: application/json' \
  -H 'cache-control: no-cache' \
  -d '{
	"ref_prev": "HEAD",
	"ref_curr": "INDEX",
	"file_name": "Untitled1.ipynb"
}'
```

(Diff changes between to Git refs in the local or remote history)

```
curl -X POST \
  http://localhost:8888/nbdime/api/gitdiff \
  -H 'Content-Type: application/json' \
  -H 'cache-control: no-cache' \
  -d '{
	"ref_prev": "origin/HEAD",
	"ref_curr": "HEAD",
	"file_name": "Untitled1.ipynb"
}'
```

### DIFF
--- a/nbdime/gitfiles.py
+++ b/nbdime/gitfiles.py
@@ -126,7 +126,7 @@ def _get_diff_entry_stream(path, blob, ref_name, repo_dir):
     return EXPLICIT_MISSING_FILE
 
 
-def changed_notebooks(ref_base, ref_remote, paths=None, repo_dir=None):
+def changed_notebooks(ref_base, ref_remote='WORKING', paths=None, repo_dir=None):
     """Iterator over all notebooks in path that has changed between the two git refs
 
     References are all valid values according to git-rev-parse. If ref_remote
@@ -144,9 +144,13 @@ def changed_notebooks(ref_base, ref_remote, paths=None, repo_dir=None):
         paths = [os.path.join(*(popped + (p,))) for p in paths]
     # Get tree for base:
     tree_base = repo.commit(ref_base).tree
-    if ref_remote is None:
-        # Diff tree against working copy:
+    if ref_remote == 'WORKING' :
+        # If a None value is provided, GitPython diffs against the Working Tree 
+        # https://gitpython.readthedocs.io/en/stable/reference.html#module-git.diff
         diff = tree_base.diff(None, paths)
+    elif ref_remote == 'INDEX':
+        # If no value is provided, GitPython diffs against the Index
+        diff = tree_base.diff(paths=paths)
     else:
         # Get remote tree and diff against base:
         tree_remote = repo.commit(ref_remote).tree
--- a/nbdime/gitfiles.py
+++ b/nbdime/gitfiles.py
@@ -93,7 +93,7 @@ def is_path_in_repo(path):
         return False
 
 
-def _get_diff_entry_stream(path, blob, ref_name, repo_dir):
+def _get_diff_entry_stream(path, blob, ref_name, repo_dir, special_ref_name):
     """Get a stream to the notebook, for a given diff entry's path and blob
 
     Returns None if path is not a Notebook file, and EXPLICIT_MISSING_FILE
@@ -101,7 +101,7 @@ def _get_diff_entry_stream(path, blob, ref_name, repo_dir):
     if path:
         if not path.endswith('.ipynb'):
             return None
-        if blob is None:
+        if special_ref_name == 'WORKING':
             # Diffing against working copy, use file on disk!
             with pushd(repo_dir):
                 # Ensure we filter if appropriate:
@@ -115,6 +115,10 @@ def _get_diff_entry_stream(path, blob, ref_name, repo_dir):
                     return io.open(path, encoding='utf-8')
                 except IOError:
                     return EXPLICIT_MISSING_FILE
+        elif blob is None:
+            # GitPython uses a None blob to indicate if a file was deleted or added.
+            # https://gitpython.readthedocs.io/en/stable/reference.html#git.diff.Diff
+            return EXPLICIT_MISSING_FILE
         else:
             # There were strange issues with passing blob data_streams around,
             # so we solve this by reading into a StringIO buffer.
@@ -133,7 +137,7 @@ def changed_notebooks(ref_base, ref_remote, paths=None, repo_dir=None, special_r
     
     - If ref_remote is None or special_remote is None, the difference is taken between
         ref_base and the working directory. (Default Case for Backwards Compatibility)
-    - If special_remote is WORKING, the difference is taken between ref_base and the working directort
+    - If special_remote is WORKING, the difference is taken between ref_base and the working directory
     - If special_remote is INDEX, the difference is taken between ref_base and the index.
 
     Iterator value is a base/remote pair of streams to Notebooks
@@ -147,13 +151,17 @@ def changed_notebooks(ref_base, ref_remote, paths=None, repo_dir=None, special_r
     if paths and popped:
         # All paths need to be prepended by popped
         paths = [os.path.join(*(popped + (p,))) for p in paths]
-    # Get tree for base:
-    tree_base = repo.commit(ref_base).tree
+
     if (not ref_remote) and (not special_remote):
-        # If a None value is provided, GitPython diffs against the Working Tree 
+        # Default case for backwards compatibility
+        special_remote = 'WORKING'
+
+    # Get tree for base and do the actual diff
+    tree_base = repo.commit(ref_base).tree
+
+    if special_remote == 'WORKING':
+        # If a None value is provided, GitPython diffs against the Working Tree
         # https://gitpython.readthedocs.io/en/stable/reference.html#module-git.diff
-        diff = tree_base.diff(None, paths)
-    elif special_remote == 'WORKING':
         diff = tree_base.diff(None, paths)
     elif special_remote == 'INDEX':
         # If no value is provided, GitPython diffs against the Index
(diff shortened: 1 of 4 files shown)


vidartf:
Thanks for this. Two small points for now:
- Would it make more sense to use objects for special refs (`ref_prev: {special: 'INDEX/WORKING'}`) to clearly differentiate, and still allow for tags to be called `INDEX` or `WORKING` (very minor corner case, but maybe more extensible).
- Would it be more explicit to rename `file_name` to `file_path`? Does it accept any path (relative and absolute)?

jaipreet-s:
> * Would it make more sense to use objects for special refs (`ref_prev: {special: 'INDEX/WORKING'}`) to clearly differentiate, and still allow for tags to be called `INDEX` or `WORKING` (very minor corner case, but maybe more extensible).

I like it. How do you feel about just making it a string with a reserved prefix such as  `special:INDEX` so that the API type is still a simple string for both regular git refs and "special" refs

vidartf:
@jaipreet-s That just renames the corner case. I figured since most of the improvement of this API over the one in my branch was by getting rid of "magic" prefixes, we should go all the way.

jaipreet-s:
All in for going all the way! My latest commit add the special case handling for the `current_ref`. I haven’t added this for the `prev_ref` yet since it didn’t quite make sense. For e.g., 

* `{curr_ref={“special”: “WORKING”}, prev_ref={“git”: “HEAD”}`
    * This makes sense since we’re trying to see the changes in the Working Tree over the HEAD ref
* `{curr_ref={“git”: “HEAD”}, prev_ref={“git”: “HEAD”}`
    * This didn’t quite make sense since it feels backwards.


For for prev_ref, the only accepted ref is “git” for now, but this is forwards compatible since we can just add a “special” field if the need arises

jaipreet-s:
> Would it be more explicit to rename file_name to file_path? Does it accept any path (relative and absolute)?

Yes this should be `file_path`. This accepts both relative paths to the Jupyter server, or absolute file paths anywhere on the file system

jaipreet-s:
The latest CI build failure is because of the JS tests, the Python tests are good. https://travis-ci.org/jupyter/nbdime/jobs/559806046

jaipreet-s:
@vidartf : Do you have a chance to take a look at this? Understood you were busy with the JL1.0 work and vacation so this did not get much traction. Thanks!

vidartf:
@jaipreet-s Thanks for the ping. I'm having a look, will push some code for discussion.

vidartf:
@jaipreet-s I cleaned it up a bit. A quick overview:
- Avoids using "special ref" extra argument in internal functions.
- Moved `get_git_notebooks` to a new base class to keep the code in `nb_server_extension` and off the more generic base.
- Renamed ref names to local/remote instead of prev/curr to keep it more in line with the rest of nbdime (although maybe we should use base/remote).
- Allows local to be "index" such that index can be diffed to working tree ("changes that *can* be staged, but which are currently not").
- Expanded the test to include some more combinations.

vidartf:
Please review and let me know any comments (probably best to comment directly on [the commit](https://github.com/jupyter/nbdime/pull/482/commits/979b7eab7ff59d82dd5835495396d7a0a689d671)?)

nbdime/gitfiles.py
```diff
@@ -93,15 +100,15 @@ def is_path_in_repo(path):
...
+def _get_diff_entry_stream(path, blob, ref_name, repo_dir):
     """Get a stream to the notebook, for a given diff entry's path and blob
 
     Returns None if path is not a Notebook file, and EXPLICIT_MISSING_FILE
     if path is missing."""
     if path:
         if not path.endswith('.ipynb'):
             return None
-        if special_ref_name == 'WORKING':
+        if blob is None:
```
jaipreet-s: There's an issue here while performing the diff between two Git refs,  where a file was added. The Base always points to the "Working Tree" version, whereas it should be empty,  so when the client computes the diff on top of the base it shows an incorrect result. I attempted to fix this in the commit https://github.com/jupyter/nbdime/pull/482/commits/ef13feb1bf511a724e437bf35b3c3bc782b0cf06. I'll try and have a stab at fixing this over your commits.

nbdime/tests/test_server_extension.py
```diff
@@ -96,26 +96,44 @@ def test_diff_api(git_repo2, server_extension_app):
...
+        ('INDEX', 'HEAD', 'diff.ipynb'),
+        ('INDEX', 'WORKING', 'sub/subfile.ipynb'),
+        ('index', 'working', 'sub/subfile.ipynb'),
+        ('iNdeX', 'WorKING', 'sub/subfile.ipynb'),
+    ):
+        print(args)
+        r = requests.post(
+            url, headers=auth_header,
+            data=json.dumps({
+                'ref_local': _make_ref(args[0]),
```
jaipreet-s: The reason I went with `prev` and `curr` was because a request like [1] to diff between the Working Tree and `origin/HEAD`, which can get a bit confusing since 'origin/HEAD' is a remote, and WORKING is actually the local changes. I'm fine with this naming if the intention is to keep things consistent.

[1] 

```
    "ref_local": {
        "git": "origin/HEAD"
        
    },
    "ref_remote": {
        "special": "WORKING"
    },
```

nbdime/gitfiles.py
vidartf: I don't really understand this issue. In the docs you link in the commit (https://gitpython.readthedocs.io/en/stable/reference.html#git.diff.Diff), it says that for added/deleted files the `path` will be `None`. We check if the `path` is `None` at the start of the function. According to the docs, if the path is not `None` and the `blob` is, then we are diffing against the working tree.

Would you mind sharing a minimal reproducible sample? I don't really understand the screenshot..

jaipreet-s: @vidartf 

Repro setup:

* Created this repo for the repo case https://github.com/jaipreet-s/diff-harness/commits/master
* I have some changes in my Working Tree, `print('Simple Diff Committed` => `print('Simple Diff in Working Tree)`
* We're trying to diff between commits eaf1741 and b82e191 in the repo so the API input is [1].

Repro output:
*  In the current state of the API (at commit https://github.com/jupyter/nbdime/commit/7bb092833ae3271c5d8fbd66e9d82cccf9362abd) the output is as expected with base file empty  [here](https://gist.github.com/jaipreet-s/396da4a2c13f476dec393e480b4e8b20#file-correct ).
*  In the state at https://github.com/jupyter/nbdime/pull/482/commits/94100ab615efad1fd388bc230306f3dd0b290cf7, the "base" file is the state in the Working Tree [here](https://gist.github.com/jaipreet-s/396da4a2c13f476dec393e480b4e8b20#file-incorrect)

So the issue is about the "base" output being returned incorrectly if we're trying to diff a commit in the history where a file has been added/

[1] 
```
{
	"ref_local": {
		"git": "eaf1741"
		
	},
	"ref_remote": {
		"git": "b82e191"
	},
	"file_path": "SimpleDiff.ipynb"
}
```

jaipreet-s: >  According to the docs, if the path is not None and the blob is, then we are diffing against the working tree.

I've documented the values for path, blob, and ref_name for 3 cases https://gist.github.com/jaipreet-s/396da4a2c13f476dec393e480b4e8b20#file-ref-blob-path-values

* _eaf1741 vs b82e191_: blob is `None` in `eaf1741` since the file was first added in `b82e191`. ref_name is the Git ref in both cases.
* _HEAD vs WORKING_: blob is `6ee0` for HEAD. For Working, ref_name **as well as** blob is None.
* _HEAD vs INDEX:_ blob is 6ee0 for HEAD. For Index, ref_name is `Diffable.Index` and blob is `58d9d9`

So, the issue is coming from the fact that `blob` can be `None` both in the Working Tree and if file is added/removed in a Git ref. But only the Working Tree will have the `ref_name` as None. 

That is why in https://github.com/jupyter/nbdime/commit/7bb092833ae3271c5d8fbd66e9d82cccf9362abd I've checked if `ref_name` is the `WorkingTree` explicitly and only then use the content from the Working Tree.

vidartf: Ok, I see now on the GitPython repo there is an issue about this https://github.com/gitpython-developers/GitPython/issues/749 .

vidartf:
Add some minor tweaks and cleanups, merging! Thanks for this!

jaipreet-s:
> Add some minor tweaks and cleanups, merging! Thanks for this!

Thanks @vidartf. Is there a plan to release a minor/pre-release version with this API so that clients (e.g. jupyterlab-git) can declare a dependency on it? (Otherwise the alternate would be a direct GitHub dependency)

vidartf:
I was originally thinking that I would try to get the visual stuff in as well, but since this only touches the Python side, I guess a minor release would be quite simple. Aiming for a release on Monday then.

jaipreet-s:
@vidartf - I tried to release a version of jupyterlab-git with a Git dependency on nbdime, but PyPI doesn't allow uploading a package with a URL dependency. So, we'll need a new minor version of nbdime to release the visual diff changes for jupyterlab-git

vidartf:
1.1.0 is out

PR #483, opened 2019-07-02
Update for lab 1.0

vidartf:
- Updates dependencies on lab to 1.0.
- Updates other web dependencies.
- Relies on lab styles now pulling in their full style dependencies. This should be double-checked carefully (that no styles break), possibly going back to the old pattern by importing `base.css` from the dependencies instead. See https://github.com/jupyterlab/jupyterlab/pull/6395.
- Some styling is now a little weird due to changes to CSS variables. I'll be looking to fix this together with #469.
- Moves codemirror to be a peer dependency in order to pick up the version that the current lab code uses.


bavaria95:
Any status on that? Need any help?

timnlupo:
Do you have any updates? Need this before I can update my extension to v1. Let me know if there's anything I can do to help

vidartf:
This update is currently blocked on some issues with the JS testing framework now that lab switched to ES2017 compilation output. If it proves to take much longer to fix that, we might do a release with manual testing only.

krassowski:
So you would need something like https://github.com/jupyterlab/jupyterlab/pull/6776 but for karma/lerna combination, right?

Just cross-referencing what might be relevant if someone has experience and would want to help out:
- https://github.com/jupyterlab/jupyterlab-git/pull/374
- https://github.com/facebook/jest/issues/3112#issuecomment-300069461 a thread for lerna/jest combination and a similar issue

vidartf:
Lerna is not a part of the issue here. The current issue is with `karma-typescript` (or possibly my configuration of it). There are many other possible ways to configure a testing framework, but karma-typescript has some clear advantages. I'd rather not throw them all out the window if it turns out all I need is a simple fix.

vidartf:
Finally got it configured (some magic involved). All set now,

bavaria95:
@vidartf, any plan to have a new release now, with support for lab 1.0?

vidartf:
@bavaria95 There was a new release yesterday 🚀 (got onto conda-forge today)

PR #492, opened 2019-08-08
fix `_compare_mimedata`

chrisjsewell:
Surprised this hasn't been noticed before. For example, this causes minor modifications of `image/svg+xml` to not be erroneously reported as deleted/inserted.


chrisjsewell:
For completeness, I also added comparison of `application/json` mimedata, using `json.dumps`

vidartf:
I included the first commit in #543. I'm not sure if the JSON one makes sense, as we already have structured JSON differ :) At the very least, I think we would need to use a sorted JSON serializer to have it be reliable.

What are the advantages to diffing JSON outputs as strings as you see it?

Carreau:
I'm doing some cleanup and this has not been touched in some time. I'm going to close; feel free to send a new PR or ping a maintainer to reopen.

PR #494, opened 2019-08-09
Honour absolute paths during git diffs

jaipreet-s:
Previously, the `get_git_notebooks` would always pre-pend the Jupyter server root to the passed path, even if it was an absolute path. This change makes it such that that is only done if an absolute path. This is helpful in cases when the Git repo root is separate from the server root.

Tested for both cases on the `/gitdiff` API by passing both `/server/repo/file` and and `repo/file`

cc @vidartf for discussion

This is one way to address https://github.com/jupyterlab/jupyterlab-git/issues/391

### DIFF
--- a/nbdime/webapp/nb_server_extension.py
+++ b/nbdime/webapp/nb_server_extension.py
@@ -87,10 +87,14 @@ class BaseGitDiffHandler(ApiDiffHandler):
         :param ref_remote: the Git ref for the "remote" or the "current" state
         :return: (base_nb, remote_nb)
         """
-        # Sometimes the root dir of the files is not cwd
-        nb_root = getattr(self.contents_manager, 'root_dir', None)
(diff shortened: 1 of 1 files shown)


vidartf:
I think I might prefer this to fixed on the client side, to be more consistent with the rest of the jupyter system. That would e.g. allow us to modify this code in the future to support other kinds of content managers as well, without breaking the API.

jaipreet-s:
For my own edification, how does this close the door on supporting other kinds of contents manager? If a relative path is passed, it still uses the `root_dir` attribute of the contents manager to resolve the absolute path to the file.

fcollonval:
This is no more required by _@jupyterlab/git_. And I support the relative path approach for consistency with the Jupyter ecosystem. So this is safe to close.

Issue #497, opened 2019-08-25
git-nbdiffdriver fails to parse branch1:file1..branch2:file2 syntax


dniku:
I came across this issue while maintaining the https://github.com/yandexdataschool/Practical_RL repository, so I will provide examples based on it.

With the following setting in `~/.gitconfig`

```
[diff "jupyternotebook"]
    command = git-nbdiffdriver diff
```

the following command fails:

```bash
$ git diff master:week07_seq2seq/practice_tf.ipynb..coursera:week6_outro/seq2seq/practice_tf.ipynb
usage: git-nbdiffdriver [-h] [--version] [--config] [--log-level {DEBUG,INFO,WARN,ERROR,CRITICAL}] {diff,webdiff,config} ...
git-nbdiffdriver: error: unrecognized arguments: index da535d8..a2c89e8 100644

fatal: external diff died, stopping at week07_seq2seq/practice_tf.ipynb
```

Apparently, either `git-nbdiffdriver` does not parse its arguments properly, or git does not pass them correctly.

`nbdiff-web` also has a similar issue:

```python
$ nbdiff-web master:week07_seq2seq/practice_tf.ipynb..coursera:week6_outro/seq2seq/practice_tf.ipynb
Traceback (most recent call last):
  File "/home/ser/.local/bin/nbdiff-web", line 10, in <module>
    sys.exit(main())
  File "/home/ser/.local/lib/python3.7/site-packages/nbdime/webapp/nbdiffweb.py", line 96, in main
    return main_diff(opts)
  File "/home/ser/.local/lib/python3.7/site-packages/nbdime/webapp/nbdiffweb.py", line 79, in main_diff
    base, remote, path = resolve_diff_args(opts)
  File "/home/ser/.local/lib/python3.7/site-packages/nbdime/args.py", line 406, in resolve_diff_args
    if not is_gitref(base):
  File "/home/ser/.local/lib/python3.7/site-packages/nbdime/gitfiles.py", line 77, in is_gitref
    is_valid_gitref(candidate)
  File "/home/ser/.local/lib/python3.7/site-packages/nbdime/gitfiles.py", line 86, in is_valid_gitref
    repo.commit(ref)
  File "/usr/lib/python3.7/site-packages/git/repo/base.py", line 466, in commit
    return self.rev_parse(text_type(rev) + "^0")
  File "/usr/lib/python3.7/site-packages/git/repo/fun.py", line 322, in rev_parse
    obj = obj[rev[start:]]
[...]

PR #498, opened 2019-08-30
Parse rename metadata CLI parameter

dniku:
When git detects that it diffs renamed files, it passes two extra
undocumented parameters to the external diff program. One of the
parameters, rename_to, has been supported since 59aef29 (2016-05-03),
but the other has been omitted. This caused (a part of) #497.
This change adds support for the missing CLI parameter.


vidartf:
Thanks for the contribution!

PR #503, opened 2019-11-18
NO-MERGE: CI check

vidartf:


vidartf:
@minrk I don't want to deal with py2 any more 😢 Thoughts on cleanly sunsetting our py2 support? :)

minrk:
> Thoughts on cleanly sunsetting our py2 support? 

No objections from me, see #504.

PR #506, opened 2020-01-04
Make read_notebook() compatible with UTF-16 files

scottcode:
Jupyter notebook files (ipynb) can be encoded in different ways. All the various `nbdime` apps break if the notebook file is not UTF-8. 

Specifically, if the notebook is encoded in UTF-16, then `nbdime.utils.read_notebook()` throws a `UnicodeDecodeError`. 

To fix this, a step was added to check the beginning of the file for a byte order mark (BOM), which would indicate UTF-16. 

A regression test for this issue was also added.


scottcode:
It looks like the test suite is peppered with opening files hard-coding `utf8` as the encoding (e.g. `nbdime/tests/conftest.py:331`). Those places would also need to be changed to make the whole repo `utf-16`-friendly.

bollwyvl:
The SHALL from the rfc:

https://tools.ietf.org/html/rfc7159

JSON text SHALL be encoded in UTF-8, UTF-16, or UTF-32.  The default
   encoding is UTF-8, and JSON texts that are encoded in UTF-8 are
   interoperable in the sense that they will be read successfully by the
   maximum number of implementations; there are many implementations
   that cannot successfully read texts in other encodings (such as
   UTF-16 and UTF-32).

   Implementations MUST NOT add a byte order mark to the beginning of a
   JSON text.  In the interests of interoperability, implementations
   that parse JSON texts MAY ignore the presence of a byte order mark
   rather than treating it as an error.


So, YES, should parse other stuff, but MUST NOT use a BOM when creating.
The MAY when reading, and encouraging non-standard writer implementations,
is likely not a thing Jupyter should be in the business of.

It's likely relevant to check what nbformat and nbconvert do, and
preferentially using an upstream's implementation. If they can't parse
them, then there is precedent for not supporting additional encodings here,
and PRs should be directed there so everybody benefits!

scottcode:
Thanks, @bollwyvl, that's very helpful context. 

So Jupyter and nbdime should in principle support notebooks saved in UTF-16, as long as they don't have BOM (as per the JSON standard). However, as `nbdime` stands, it does not currently support UTF-16 even without BOM, because in the code (core and tests) it explicitly specifies UTF-8 everywhere. 

The weird thing is I'm not sure what caused some of my notebooks to end up in UTF-16 to begin with. I was just using Jupyter like normal. Since those notebooks were working in Jupyter, it appears that Jupyter supports UTF-16 encodings. Since Jupyter supports them, should we consider modifying `nbdime` to make it compatible (assuming we don't worry about supporting BOM)?

scottcode:
What happens if we just leave off all the explicit encodings? Wouldn’t the open() statements just infer the right encoding?

vidartf:
As we are simply using `nbformat.read` internally, I think this should be redone as an issue on the nbformat repo for discussion, as the behaviour should be consistent across all the Jupyter landscape :)

scottcode:
Sounds reasonable, @vidartf. Thanks for your consideration and for the suggestion.

PR #551, opened 2020-11-03
Update to Jupyterlab 3.0

ajbozarth:
Updates dependancies to Jupyterlab 3.0 rc6 and preemptively increments a major version to allow for testing with jupyterlab-git extension, sister PR at https://github.com/jupyterlab/jupyterlab-git/pull/818

Marked as draft until 3.0 release when this PR will be updated.


ajbozarth:
@lresende

ajbozarth:
@vidartf I just updated this and tested that it works with rc13. Is there any way we could merge this as is and release an rc? That would allow for easier review and editing of https://github.com/jupyterlab/jupyterlab-git/pull/818 which depends on this update.

vidartf:
@ajbozarth Thanks for this. A few questions:
- It seems to me as if this PR will cause nbdime to no longer be available as a server extension for jupyter notebook. As there are likely still a decent amount of users on that platform, it would be nice to still offer it. Do you know if there is a way to retain support for that while also making it available as a jupyter server extension? (cc @Zsailer )
-  Will this PR also add support for nbdime being used as a federated extension? I'm happy to do that in a separate step, but I assume from the diff that it does not ?

ajbozarth:
@vidartf 

> It seems to me as if this PR will cause nbdime to no longer be available as a server extension for jupyter notebook. As there are likely still a decent amount of users on that platform, it would be nice to still offer it. Do you know if there is a way to retain support for that while also making it available as a jupyter server extension? (cc @Zsailer )

Thanks for reminding me of this, it was unintentional, I was unable to get the notebook extension working on 3.0 so I just updated it to a server extension to fix it. @Zsailer is there a way to support both at once? if not I'll try to roll this back and continue to debug why the notebook extension isn't working on 3.0

> Will this PR also add support for nbdime being used as a ~~federated~~ pre-built extension? I'm happy to do that in a separate step, but I assume from the diff that it does not ?

This is intentional my part, as I mentioned in a dev meeting I'm focusing on updating extensions first to work on 3.0 "as is" for a quick turn-around release to unblock downstream dependencies then follow up later with any updates to to the new pre-build extensions

Zsailer:
Yup, you can get an extension to work both in classic notebook and jupyter_server.

[These docs](https://jupyter-server.readthedocs.io/en/latest/developers/extensions.html#migrating-an-extension-to-use-jupyter-server) should be helpful in Jupyter Server.

Zsailer:
As a side note, I tested nbdime against jupyter_server + classic notebook here: https://github.com/Zsailer/jpserver-extension-check

nbdime should work as-is thanks to nbclassic (though it's not a bad idea to update things).

ajbozarth:
@Zsailer thank you, I'll follow up on this later today

ajbozarth:
I've pushed an update that should reenable the notebook server extension for nbclassic, but as far as I can tell only the jupyter server extension works with 3.0

@Zsailer that link to the check you shared only checks if the extension is successfully installed, which it is. What I'm seeing when I only install the `serverextension` and not the `server extension` is I get 404s from the nbdime API. We saw the same issue when upgrading Elyra, in fact I've personally yet to see a `serverextension` not give 404s on lab 3.0

vidartf:
@ajbozarth / @Zsailer Any news on the server extension parts? Please note that there are unittests for the server extensions that are all failing right now.

ajbozarth:
We figured out the build issues and I'm updating the docs now. I'll take a look at the tests this afternoon once I finish the docs.

ajbozarth:
Sorry this stalled, I have wip doc and tests for this ready as of late last Friday but haven't had time to finish it up since my holiday vacation started. I'm hoping to take a hour or so sometime when I have a chance to get those updates finished and pushed. Should find time before New Years, hopefully before Christmas.

vidartf:
As noted in https://github.com/jupyter/nbdime/pull/552#issuecomment-748090429, we will also need to figure out how to get MathJax onto the page for typsetting equations, as it will no longer be available via the notebook package.

ajbozarth:
I am back from holiday and working on this again, I should have new updates pushed by EOD

ajbozarth:
@vidartf This is ready to merge for a rc release. We still need to figure out what's causing the CI failure, but the tests only seem to fail on CI and not locally. Since this is blocking development of jupyterlab/jupyterlab-git#818 it would be better to merge, cut the rc, then address the tests in a follow up PR prior to a final release.

vidartf:
I'm getting failures locally as well:
```python
file <repo root>\nbdime\tests\test_jupyter_server_extension.py, line 55
  @pytest.mark.timeout(timeout=WEB_TEST_TIMEOUT)
  async def test_isgit(git_repo2, jp_fetch):
      r = await jp_fetch(
          'nbdime/api/isgit',
          method='POST',
          body=json.dumps({
              'path': git_repo2,
          })
      )
      code, body = read_response(r)
      assert body == {'is_git': True}
file <env root>\lib\site-packages\jupyter_server\pytest_plugin.py, line 324
  @pytest.fixture
  def jp_fetch(jp_serverapp, http_server_client, jp_auth_header, jp_base_url):
file <env root>\lib\site-packages\jupyter_server\pytest_plugin.py, line 291
  @pytest.fixture(scope="function")
  def jp_serverapp(
file <env root>\lib\site-packages\jupyter_server\pytest_plugin.py, line 199
  @pytest.fixture(scope='function')
  def jp_configurable_serverapp(
file <env root>\lib\site-packages\jupyter_server\pytest_plugin.py, line 173
  @pytest.fixture
  def jp_http_port(http_server_port):
E       fixture 'http_server_port' not found
```

Seems to be an issue with the `jupyter_server` pytest plugin (missing dependency? wrong version of a dependency?)

vidartf:
It seems that jupyter-server's dependency `pytest-tornasync` conflicts with nbdime's dependency `pytest-tornado` (they share some fixture names).

vidartf:
I've pushed a commit that fixes the tests. There is currently only a single test failing, the "offline mathjax" test when used as an extension for jupyter server. I'm currently planning to create a separate jupyter server extension that will expose these resources, and have nbdime depend on this. I'll try to do a final review on this PR, merge and do a pre-release this week, and then open a separate issue/PR to resolve the MathJax issue as a blocker for a final release.

nbdime/__init__.py
```diff
@@ -18,12 +18,18 @@ def load_jupyter_server_extension(nb_server_app):
...
+_load_jupyter_server_extension = load_jupyter_server_extension
+
+
 def _jupyter_server_extension_paths():
     return [{
         "module": "nbdime"
     }]
 
 
+_jupyter_server_extension_points = _jupyter_server_extension_paths
```
vidartf, line 30: This is currently not included in `__all__`. Should it be?

nbdime/webapp/nb_server_extension.py
```diff
@@ -8,9 +8,15 @@
...
-from notebook.utils import url_path_join, to_os_path
-from notebook.services.contents.checkpoints import GenericCheckpointsMixin
-from notebook.services.contents.filecheckpoints import FileCheckpoints
+from jupyter_server.utils import url_path_join, to_os_path
+
+from jupyter_server.services.contents.checkpoints import GenericCheckpointsMixin as jpserver_GenericCheckpointsMixin
+from jupyter_server.services.contents.filecheckpoints import FileCheckpoints as jpserver_FileCheckpoints
+
+from notebook.services.contents.checkpoints import GenericCheckpointsMixin as nbserver_GenericCheckpointsMixin
+from notebook.services.contents.filecheckpoints import FileCheckpoints as nbserver_FileCheckpoints
```
vidartf, line 17: I'll try to make these imports guarded so we can have optional dependencies here (and drop the `notebook` dependency from setup.py).

packages/nbdime/package.json
```diff
@@ -23,22 +23,22 @@
...
-    "@lumino/dragdrop": "^1.3.0",
-    "@lumino/signaling": "^1.2.2",
-    "@lumino/widgets": "^1.6.0",
+    "@jupyterlab/codeeditor": "^3.0.0",
+    "@jupyterlab/codemirror": "^3.0.0",
+    "@jupyterlab/coreutils": "^5.0.0",
+    "@jupyterlab/nbformat": "^3.0.0",
+    "@jupyterlab/outputarea": "^3.0.0",
+    "@jupyterlab/rendermime": "^3.0.0",
+    "@jupyterlab/services": "^6.0.0",
```
vidartf, line 32: Given that none of the JS code needed to change, do you think we should change the dependencies to `^5 || ^6` (and similar for other packages)? That would allow us to still support lab2 for now.

vidartf (commented):
Most looks good now, thanks for the effort! 👍 I have some minor points for input below, but nothing that should be a lot of work.

nbdime/__init__.py
ajbozarth: `_jupyter_server_extension_paths` wasn't in `__all__` so I didn't add the mask function `_jupyter_server_extension_points` either. If you think they should be added I can do that.

packages/nbdime/package.json
ajbozarth: I believe that would cause issues due to other dependancies like Lumino which need to run different versions depending on the lab version. Plus I would expect we would follow up this initial release with an update to pre-built extensions, which wouldn't be backward compatible. In general I think having user use nbdime 2.X for Lab 2.X and nbdime 3.X for lab 3.X make sense (to me)

ajbozarth:
@vidartf do you think we're close to merging this and cutting an rc? It would help enable review of the down stream git-extension PR

vidartf:
I'm working on fixing the JS tests, but if I can't fix them by EOD tomorrow, I'll try to just release an RC as-is.

vidartf:
Turns out the new nbformat release also breaks CI (cf #553 ), but I'll cut a beta.

vidartf:
I've done beta releases base off of https://github.com/jupyter/nbdime/commit/c82362344e596efdc4f54c927d90338940e0fa41

ajbozarth:
Do we have a timeline on a final release of 3.0?

PR #552, opened 2020-11-22
Add nbdiff-web-export to export HTML diff using just command line

trams:
This is Proof-of-concept but it should be good enough to actually try to use it

How to try it?
    nbdiff ./nbdime/webapp/testnotebooks/remote.ipynb ./nbdime/webapp/testnotebooks/base.ipynb --out diff_with_base.json --include-base
    nbdiff-web-export diff_with_base.json  ./nbdime/webapp/static/nbdime.js > ./diff.html
    
    Open diff.html in your browser of choice

What has been tested?
   Only the commands above

I am thinking about actually combining two command and teaching nbdiff-web-export to build diff and then html. I do not see the reason why I won't be able to import appropriate packages and call some kind of diff method. What do you think about this?

I have no idea how this would work with several changed notebooks in some git commit. I honestly did not try it. If someone can try it I would appreciate

As you can see in code change I am trying to reuse as much as possible existing components: nbdime.js to render the diff. I copy pasted and modified a template (would be nice to just reuse it but yet to figure this out). Added python code is mostly trivial

What is missing?
  1. Any kind of tests
  2. Proper error processing

I would really appreciate your feedback


trams:
I figured out how to combine two commands i.e. do the diff and then generate html to render it so I will update this pull request

nbdime/webapp/templates/diffembedded.html
```diff
@@ -0,0 +1,38 @@
...
+      <div id=nbdime-header-banner>
+        <span id="nbdime-header-base">Base</span>
+        <span id="nbdime-header-remote">Remote</span>
+      </div>
+    </div> <!-- ndime-header -->
+
+    <div id="nbdime-root" class="nbdime-root">
+    </div>
+
+    <script src="{{ nbdime_url }}"></script>
```
vidartf, line 35: You could also put the content of the script between the `<script>` tags to make the exported file truly self-contained. Or maybe have that be configurable (some flag or value for the url causes the script to be fully embedded)?

vidartf (commented):
This mostly looks good. I'm sure there is some tidying up that I might want to do, but I can do that iteratively after merging. The main concepts look good, so thanks! I have one conceptual question inline (keeping JS separate vs embedding fully).

nbdime/webapp/templates/diffembedded.html
trams: I was very reluctant to embed nbdime.js cause it is a big file (3.2Mb) and I had a case of CI/CD in mind where I would have ~1K diffs to store. So it seemed like a waste

After your comment I realized it is just *premature* optimization

Also I realized there is a legit use case when one want to render diff as HTML and sent it over Slack or something like that

So my suggestion.
I will try to make it embedded by default. If someone is providing `url` via a command line I would use it.
How does it sound?

nbdime/tests/test_cli_apps.py
```diff
@@ -286,6 +286,16 @@ def test_nbdiff_app_no_colors(filespath, capsys):
     assert 0 == main_diff(args)
 
 
+def test_nbdiff_app_fail_if_file_is_missing(filespath):
+    # Simply check that the --color-words argument is accepted, not behavior
```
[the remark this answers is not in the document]
trams: Good catch. I just copy pasted the test with a comment

nbdime/webapp/templates/diffembedded.html
trams: I updated the pull request.
Now by default the tool will embed the script. If url will be provided then the script won't be embedded

trams:
So things which are not working
1. Export button does not work (do we need it?)
2. Hide changes button does not work
3. The title of the page is generic and should contain the names of notebooks changed.

But I do believe it is already usable. I will definitely try to integrate this tool into our ML CI/CD so I will get a better sense how to make it more user friendly

trams:
@vidartf I addressed your concerns. Let me know if there is anything else preventing this from merging? Thank you

vidartf:
Oh, I thought your previous comment meant you wanted to go away and try it out some more before coming back and making adjustments. I'll try to have a look this week 👍

vidartf:
Hi again, I did some cleanup to align it more with some internal conventions, fixing some issues while doing some. There are a few remaining issues:
- Ensure we build and use a single static bundle. Existing nbdime.js does not bundle all the things (fonts, and some other chunked JS). This probably means we need another webpack config, and so we need to decide if it should it be an additional one, or modify the existing one. If we add another, it means the PyPI package will become larger, if we modify the existing one, it means the base bundle that always gets loaded will increase in size by quite a bit.
- Find and check a decent default for MathJax: Use CDN as default, or switch to using jupyter-renderer's mathjax3 bundled ? If we use a CDN as default, it means we cannot get MathJax rendering when offline. If we bundle MathJax, the bundle size will increase as well. This question ties in both to the point above (separate bundle or additional one), as well as the upgrade in #551 to switch to jupyter_server: since jupyter server does not include mathjax, we will have to make a plan for how do get mathjax onto the page for typesetting latex anyways.

vidartf:
Oh, and we will at some points need to add test coverage for the new CLI entry point.

vidartf:
An option to solve the MathJax is outline here: https://github.com/jupyter/nbdime/pull/551#issuecomment-756087493

> I'm currently planning to create a separate jupyter server extension that will expose these resources, and have nbdime depend on this.

trams:
Thank you for all the great feedback. I still did not have a chance to fully go though comments and I do not fully understand all the comments yet.

Meanwhile I had some success integrating this version in our CI/CD and it seems it works fine but to be fair in the last few weeks we did not have a lot of changes to our notebooks so we probably did not hit any edge cases

As for whether to extend the current bundle or to create a separate I do not have specific opinion.

icankeep:
hello, when the code can be merged? I really need this feature！！

wenindoubt:
@trams Thanks so much for starting the work on this. This is exactly what we need as well. Are there any updates on getting this merged?

trams:
@wenindoubt I did not have a chance to address the maintainer's comments mostly due to lack of expertise (I am a backend developer) and lack of time.
Currently I am trying to get some "buy in" on such a feature inside my company and hopefully we will allocate some time on merging this pull request. But I can't make any promises as of now

I am open on accepting anyone's help! I am open on allocating some time on zoom calls for XP sessions or consultations if needed

Carreau:
This has not been touched in a few year and I'm doing some cleanup, so I'm going to close, feel free to resubmit as a separate PR or ask a maintainer to reopen.

PR #566, opened 2021-02-27
Add initial cell identifiers awareness

### LINKED ISSUE #553: Support cell IDs
See recent JEP and implementation:
- JEP:https://github.com/jupyter/enhancement-proposals/blob/master/62-cell-id/cell-id.md
- nbformat implementation: https://github.com/jupyter/nbformat/pull/189/files?file-filters%5B%5D=.py&file-filters%5B%5D=.rst

For nbdime, we then need to consider whether we should use IDs as a high level indicator of cell identity when diffing. Questions that need to be resolved are: 
- Should IDs take precedence over content? Just using IDs would make some things a lot simpler, but as the content is the only attribute that is exposed to the user, content equality should probably still be the strongest measure?
- If IDs are not the highest precedence, should we display changed IDs? If so, how (terminal seems straight forward, web less so).
- If IDs are not the highest precedence, how do we resolve potential ID conflicts when merging? Some applications are likely to want to associate content with cells via their IDs, so if they conflict, we are unlikely to be able to auto-resolve. We would also need to figure out how to display such conflicts.

krassowski:
To keep the conversation from https://github.com/jupyter/nbdime/issues/553 going, in particular the step 1:

> step 1, ignore cell id and make a release with unchanged behavior that is aware of cell ids, as if it were any other opaque metadata field. Since cell IDs are hidden from the user, I think showing them in diffs should also be done at the changed-metadata level, maybe hidden by default.

I am opening a draft PR that might help with conceptualizing what is needed to move forward. The number of failing tests is down from 34 to 32 (fixed `test_pretty_print_code_cell` and `test_git_diff_driver_add_helper_filter` I _think_) but the bulk of merge tests is still failing. They are now failing for a different reason (I temporarily set `"/cells/*/id": "remove"` while awaiting feedback to highlight the file where the changes are likely needed).

I really wish we could get nbdime out of beta to also allow jupyterlab-git to be properly released and for the lab ecosystem users to migrate to 3.0 and reduce the maintenance burden on the older versions of other Jupyter software too.

This PR might help to close #553.


nbdime/merging/notebooks.py
```diff
@@ -58,6 +58,7 @@
 
 def notebook_merge_strategies(args):
     strategies = Strategies({
+        "/cells/*/id": "remove",
```
krassowski, line 61: Seeking maintainers input: what to do here.

vidartf: I think `remove` would be correct action for "step 1", but we would need to verify that it generates a new id before saving.

nbdime/prettyprint.py
```diff
@@ -112,6 +112,8 @@ def should_ignore_path(self, path):
             return not self.attachments
         if starred.startswith('/cells/*/metadata') or starred.startswith('/metadata'):
             return not self.metadata
+        if starred.startswith('/id'):
+            return not self.id
```
vidartf, line 116: I'm not sure if this makes sense. The path would need to be `/cells/*/id` for it to match cell ids.

vidartf (commented):
Thanks for picking up the slack here. I have some changes locally that complements this (test changes only). I'll do a PR to your branch with them.

nbdime/prettyprint.py
krassowski: Indeed, this was wrong. I changed it to` /cells/*/id`.

krassowski:
All tests that were failing due to introduction of cell ids are now passing. There are still failing tests, but unrelated:

- `test_git_difftool[jupyter_server]`:   `{'mathjaxUrl': ''} != {'mathjaxUrl': '/static/components/MathJax/MathJax.js'}`
- `test_offline_mathjax`: `tornado.httpclient.HTTPClientError: HTTP 404: Not Found`

krassowski:
Comparing [jupyter_server/search?q=mathjax_url](https://github.com/jupyter-server/jupyter_server/search?q=mathjax_url) and [notebook/search?q=mathjax_url](https://github.com/jupyter/notebook/search?q=mathjax_url) it looks that:
- the [handler](https://github.com/jupyter-server/jupyter_server/blob/3135bd6f764210cf1d084cfec3cb8ad218a4340a/jupyter_server/base/handlers.py#L217-L222
) was kept in the `jupyter_server`, but
- the actual mathjax javascript was removed from `jupyter_server` and is only present in `notebook`: [notebook/static/base/js/mathjaxutils.js](https://github.com/jupyter/notebook/blob/2ba296039aa70afbfd90630fea14bb8461b40410/notebook/static/base/js/mathjaxutils.js)
- the mathjax config was also removed from `jupyter_server` and is only present in `notebook`: [notebook/notebookapp.py#L1344-L1365](https://github.com/jupyter/notebook/blob/dc8b70519f61540cf3869ccadbff3c47d355c7b2/notebook/notebookapp.py#L1344-L1365)

vidartf:
Thanks!

PR #568, opened 2021-03-03
Add pass-by-value API server

### LINKED ISSUE #549: Feature: pass-by-value API
I want to host nbdime as a multi-tenant, stateless REST service. However the diff & merge APIs currently only support local paths or URLs as arguments. This requires either uploading the files to the server or hosting them publicly, both of which I want to avoid.

### Proposed change:
Allow the diff & merge apis to take notebook json as arguments; i.e. for `/api/diff` the request JSON would now be:
```js
{
    "base":   "filename.ipynb" | "http://your-domain/url/path" | json_notebook,
    "remote": "filename.ipynb" | "http://your-domain/url/path" | json_notebook
}
```

If the endpoint receives a dictionary for any of the arguments, it would interpret it as a notebook and use that directly.

Some kind of `--disallow-notebook-paths` server flag to disable passing a path or URL to the API would be good too, for security purposes.

I'm willing to make the changes for this, just looking for any feedback first.

kycutler:
Resolves #549

Also adds `--num-processes` argument for all server commands (Unix only).

Doesn't currently contain tests, though I've done a fair bit of testing on my own -- would appreciate any feedback for how to make automated tests work since pytest-tornado only seems to support one entry point.

I'm also open to other suggestions for what to name the new entry point -- I know "api" is a little vague.

### DIFF
--- a/nbdime/__main__.py
+++ b/nbdime/__main__.py
@@ -83,6 +83,8 @@ def main_dispatch(args=None):
         main = main_mergetool
     elif cmd == 'server':
         from nbdime.webapp.nbdimeserver import main
+    elif cmd == 'api':
+        from nbdime.webapp.nbapiserver import main
     elif cmd == 'config-git':
         # Call all git configs
         from nbdime.vcs.git.diffdriver import main as diff_driver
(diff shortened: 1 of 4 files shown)


krassowski:
I don't know if you are still interested, but it looks like this picked up some conflicts.

PR #572, opened 2021-03-13
Package labextension as federated extension

fcollonval:


packages/labextension/package.json
```diff
@@ -30,9 +30,9 @@
   "scripts": {
     "build": "npm run build:lib && npm run build:labextension",
     "build:dev": "npm run build:lib",
-    "build:labextension": "rimraf dist && mkdirp dist && cd dist && npm pack ..",
+    "build:labextension": "rimraf dist && mkdirp dist && jupyter labextension build . && cd dist && npm pack ..",
```
vidartf, line 33: The `jupyter` command requires the python jupyterlab package to be available when the command is run. To ensure this we would probably need to use PEP517 build/install flow.

fcollonval: Thanks, I added a `pyproject.toml` file.

setup.py
```diff
@@ -7,10 +7,16 @@
...
 import os
+imoprt sys
 from glob import glob
 
 from setuptools import setup, find_packages
 
+# ensure the current directory is on sys.path
+# so setupbase can be imported when pip uses
+# PEP 517/518 build rules.
+sys.path.append(os.path.dirname(__file__))
```
fcollonval, line 18: Apply the same fix than for versioneer there: https://github.com/dask/dask-labextension/pull/170/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7

We should maybe switch to `jupyter-packaging` rather than a local `setupbase` file?

fcollonval:
There is one thing missing in this PR is the definition of `_jupyter_labextension_paths`
as for example in the template: https://github.com/jupyterlab/extension-cookiecutter-ts/blob/a161a92deee694d42f52a9769ff76bacc969175f/%7B%7Bcookiecutter.python_name%7D%7D/%7B%7Bcookiecutter.python_name%7D%7D/__init__.py#L12

From my understanding it is purely to get some metadata, so should we add it. And if yes, is it okay to place the output of `nbdime-jupyterlab` directly within the python package?

fcollonval:
I modify the PR to place the JupyterLab assets within the python package.

setup.py
vidartf: Yes, but we would need to check carefully, cause I'm wondering if there might have been some drift between the two.

packages/labextension/package.json
```diff
@@ -29,10 +29,12 @@
   "typings": "lib/index.d.ts",
   "scripts": {
     "build": "npm run build:lib && npm run build:labextension",
-    "build:dev": "npm run build:lib",
-    "build:labextension": "rimraf dist && mkdirp dist && cd dist && npm pack ..",
+    "build:dev": "npm run build:lib && jupyter labextension build --development True .",
+    "build:labextension": "rimraf ../../nbdime/labextension && mkdirp ../../nbdime/labextension/federated && jupyter labextension build . && cd ../../nbdime/labextension && npm pack ../../packages/labextension",
```
vidartf, line 33: The non-federated version of the extension shouldn't need to be disted in the python package (it cannot be picked up in any way), so we can save quite a few MB by not putting it in.

vidartf (approved):
Looks good, thanks for the work! I've left two comments. I'll deal with this before releasing, unless you have any disagreements with the comments.

PR #600, opened 2021-08-30
Chore: use jupyter packaging

agoose77:
This PR started as from a switch to use jupyter-packaging to a switch to hatch-jupyter-hatch and use of modern configuration through `pyproject.toml`.


fcollonval:
The content of the wheel and the sdist have been compared to the master branch to check it is not impacted (except for the labextension that is not stored in the python source in the wheel to avoid useless duplication - aka it is stored as shared data).

PR #605, opened 2021-09-30
Format note and code correctly on installation page

joelostblom:
This section currently looks a bit odd on the installation page, so I suggest moving the note on top and format it as an admonition.

Current appearance:

image


vidartf (approved):

vidartf:
Thanks!

Issue #240, opened 2017-01-16
Pretty print color contrasts


vidartf:
In the pretty printed output, the headers are printed in blue. In terminals with black backgrounds, this leads to very poor contrast (blue on black). The green/red/yellow outputs are all OK (at least to my eyes). We should consider if there is another color that would fit better, but that would also work for terminals with light background.

PR #615, opened 2022-02-04
Make INFO text more readable in black terminal

tallamjr:
As addressed in https://github.com/jupyter/nbdime/issues/240 the
headers are printed in BLUE, which is hard to read when using git
integration in a dark terminal.

This allows for the color to remain the same, but just to brighten it up
such that is more readable.

Refs:
    - https://github.com/tartley/colorama/blob/34703ece90fc4996727adff31502bad8a8e88066/demos/demo03.py#L15

	modified:   prettyprint.py

### DIFF
--- a/nbdime/prettyprint.py
+++ b/nbdime/prettyprint.py
@@ -57,7 +57,7 @@ col_const = {
         KEEP   = '{color}   '.format(color=''),
         REMOVE = '{color}-  '.format(color=colorama.Fore.RED),
         ADD    = '{color}+  '.format(color=colorama.Fore.GREEN),
-        INFO   = '{color}## '.format(color=colorama.Fore.BLUE),
+        INFO   = '{color}## '.format(color=colorama.Fore.BLUE + colorama.Style.BRIGHT),
         RESET  = colorama.Style.RESET_ALL,
     ),
 


vidartf:
Hi! How does this look on terminals with bright backgrounds? Could you post some screenshots? Preferably also stating which platform/terminal the screenshots were taken on, as I know the colors used tend to differ.

tallamjr:
> Hi! How does this look on terminals with bright backgrounds? Could you post some screenshots? Preferably also stating which platform/terminal the screenshots were taken on, as I know the colors used tend to differ.

Hi @vidartf , sure, please see below the difference without (left) and with (right) the change:

**Dark Background**
Screenshot 2022-02-09 at 14 31 23

**Light Background**
Screenshot 2022-02-09 at 14 30 07

I am running on `macOS.10.15.7` and the above terminal app is `iterm2:v3.3.11`. Unfortunately I am unable to test on other platforms since my `Terminal.app` on my mac seems to be busted and I do not have access to a linux box or windows machine.

Hope that is still useful.

vidartf:
Can confirm improvement on Windows as well:

image

and in bright theme in Windows Terminal:

image

PR #628, opened 2022-10-14
Refactor

gwincr11:
👋 Hello @vidartf 

I started a small refactor to try to get my feet wet with NbDime. I am having a difficult time understanding how to do source control on the packages and how to use lerna. Any help here would be greatly appreciated.

This pr should only change 2 files, the notebook widget and the associated typescript, but the gitignore is not tracking those anymore?

I am also wondering if you could add some docs around working with Lerna and contributing to these pieces to help us out. 

Thank you ❤️


gwincr11:
Oh I see how things are organized...

PR #630, opened 2022-10-18
chore: modernise js setup

maraisr:
Goes through some moments to moderniser our JS dev infrastructure. Most notably the removal of lerna in favour of built-in features of pnpm. There is a journey here, but this is the first step for us.

1. `lerna` + `npm` -> `pnpm`
2. `ts-loader` -> `swc` better perf, and better out-of-the-box support
3. dev env uses node 16 LTS

Eventually it'll be nice not to rely on `shamefully-hoist=true`, but for today we need this.

Up next 👉 prettier + linters.


.gitignore
```diff
@@ -29,3 +29,4 @@ package-lock.json
 *.tsbuildinfo
 packages/webapp/webpack-stats.json
 nbdime/labextension/
+/src/
```
maraisr, line 32: running `pip install -e git+https://github.com/jupyter/nbdime#egg=nbdime` adds this folder, but probably shouldn't be included in VCS — lets git ignore it

.npmrc
```diff
@@ -0,0 +1,2 @@
+strict-peer-dependencies=false
+shamefully-hoist=true
```
maraisr, line 2: lerna used to also hoist, think its due to the copying we do of our packages — ideally in a future moment we won't need to do this.

.github/workflows/tests.yml
```diff
@@ -34,45 +37,48 @@ jobs:
...
       with:
-        node-version: 12.x
-    - name: Get npm cache directory
-      id: npm-cache-dir
-      run: |
-        echo "::set-output name=dir::$(npm config get cache)"
-    - uses: actions/cache@v2
-      id: npm-cache
+        node-version: 16.x
+        cache: 'pnpm'
```
maraisr, line 63: built-in caching, lets just use that 🤝

packages/webapp/webpack.config.js
```diff
@@ -4,31 +4,97 @@ module.exports = {
...
-      { test: /\.(jpg|png|gif)$/, loader: 'file-loader' },
-      // required to load font-awesome
-      { test: /\.woff2(\?v=\d+\.\d+\.\d+)?$/, loader: 'url-loader?limit=10000&mimetype=application/font-woff' },
-      { test: /\.woff(\?v=\d+\.\d+\.\d+)?$/, loader: 'url-loader?limit=10000&mimetype=application/font-woff' },
-      { test: /\.ttf(\?v=\d+\.\d+\.\d+)?$/, loader: 'url-loader?limit=10000&mimetype=application/octet-stream' },
-      { test: /\.eot(\?v=\d+\.\d+\.\d+)?$/, loader: 'file-loader' },
-      { test: /\.svg(\?v=\d+\.\d+\.\d+)?$/, loader: 'url-loader?limit=10000&mimetype=image/svg+xml' }
+      {
+        test: /\.css$/,
+        use: [require.resolve('style-loader'), require.resolve('css-loader')],
```
maraisr, line 16: webpack will always resolve loaders relative to itself, rather than relative to the config file — better to resolve based on the config files — as that is where our deps are defined.

vidartf:
Would it make sense to add a `workspaces` entry in the root dir package.json? https://docs.npmjs.com/cli/v7/using-npm/workspaces That would also allow easier change between npm, pnpm and yarn.

The main thing I would want to verify when switching to pnpm is that any changes to the logic of deduplication of dependencies doesn't change in such a way that it affects the built bundle sizes.

maraisr:
hi @vidartf no, nothing changes by way of bundles — no third party dependencies are bundled and are at the mercy of the other parties package manager.

this change is solely for this project. 

Ideally we shouldn't optimise the project for multiple developer-centric package managers, that is just a recipe for disaster. We need to pick one, and stick to that one.

vidartf:
> hi @vidartf no, nothing changes by way of bundles — no third party dependencies are bundled and are at the mercy of the other parties package manager.
> 
> this change is solely for this project.

I mean the bundles produced by the lab extension and webapp packages (webpack). I.e. if this change causes the bundle sizes to go up for those projects due to inefficient deduplication, that would be an issue.

> Ideally we shouldn't optimise the project for multiple developer-centric package managers, that is just a recipe for disaster. We need to pick one, and stick to that one.

Right, but using `workspaces` is a standard pattern across npm dev, so it is more easily understandable by others coming into the project. It just simply didn't exist as a standard when nbdime was first configured. In that vein, yarn or vanilla npm + lerna would probably be the ones that align the most with other Jupyter ecosystem repositories, but pnpm should be fine as well, as long as it isn't too different on a conceptual basis.

maraisr:
pnpm isn't a new kid on the block either, its very popular as well. I use it all over my open source libraries, was just natural for me to gravitate towards it.

but you do what you want to do.

in terms of duplication, no there isn't any of that — as long as peer dependencies line up we'll be fine. But that is the same problem you'll get with yarn and or npm as well.

In anycase, don't think we should get too hung up on the package manager, pnpm just had a direct replacement for what lerna offered. As well as performance gains, and so forth. `workspaces` configuration item is the same across all managers, its just where its defined. Yarn/npm its in the root package json, bolt its in a bolt sub-field, pnpm is in a new file.

Let me know what you want to do; we need to uplift the js hygiene in this project — the package manager was just the first step.

maraisr:
Ive got a pnpmless code here, I'll need to get it to work, feel the build topology isn't quite right, but let me know.

https://github.com/maraisr/nbdime/commit/2c6124d7de1d2dffe7f330995aaa27204c032c12

let me know where to go from here 😅 🤝

---

I know bundle size is important to you (and I actually), and from what I can tell we're shipping regenerator-runtime and core-js with the babel derivative (which are both huge), ill do a proper bundle analysis for us — but believe that we can achieve much better bundles outside of which package manager we use.

PR #632, opened 2022-10-18
chore: upgrade webpack to version 5

maraisr:
to be merged after #631 

This PR upgrades webpack to 5


maraisr:
quite possible this pr breaks images, will circle back

maraisr:
quite possible that this pr breaks images, will need to confirm

Issue #635, opened 2022-10-20
Lazy Cell Display For Unchanged Cells


gwincr11:
Rendering unchanged cells is a very high overhead process for NbDime. Currently all the cells are rendered and then hidden with css. 

We are going to open a pull request to introduce a new method of Lazy Loading the unchanged cells only when needed. This will work in an expand up and down one cell at a time process. 

We currently have this working in our nbdime spike on GitHub and once we have it well tested hope to upstream it here. 

Before expansion
<img width="1619" alt="Screen Shot 2022-10-19 at 9 35 00 PM" src="https://user-images.githubusercontent.com/289882/196835451-61f11ccb-a0df-4c01-8776-e40003d7af3c.png">

Expanded up
<img width="1615" alt="Screen Shot 2022-10-19 at 9 35 10 PM" src="https://user-images.githubusercontent.com/289882/196835465-e41a6b45-21ef-4b6c-a525-a2bb19fea9ce.png">

vidartf:
This seems very sensible. 👍

gwincr11:
Great, I will work on upstreaming this code. I am also going to take on a few small refactors as well, I will do them in separate PR's

gwincr11:
I have started working on upstreaming what we have built, it may take a bit of adjustment to make it work with the setup in here versus our typescript setup. 

I would love to get some of @maraisr's work in so I don't have to fix a bunch of linter and typescript differences if those are being seriously considered.

PR #636, opened 2022-10-24
Make Cells Display Lazily When Diffing Unchanged Cells

gwincr11:
Motivation:
  During diffing Unchanged cells are displayed, which is not necessary.
  This is a performance issue when there are many unchanged cells.

Related Issues:
  - https://github.com/jupyter/nbdime/issues/635

Open Questions:
  - I am struggling with getting the svg images to build properly.
  - How can this be tested in JupyterLab?

Changes:
  - Alter Notebooks Widget to only display changed cells by using new linked list of cells.
  - Add new linked list of cells and lazy version of linked list of cells.


gwincr11:
@vidartf any hints on how to get the SVG to render properly?

vidartf:
> @vidartf any hints on how to get the SVG to render properly? If not I can inline it 😅

I'm unsure what the best approach is here. I think if we want import to work, it will be a breaking change, as any library that depends on nbdime will now have to figure out bundling rules for SVGs. I think also some care need to be taken about adding classes etc to the SVG elements, so that JupyterLab can theme it properly. I think it might be good to read up on https://jupyterlab.readthedocs.io/en/stable/extension/ui_components.html#how-to-create-your-own-custom-labicon

packages/nbdime/src/styles/variables.css
```diff
@@ -27,4 +27,10 @@
 
   --jp-merge-either-color1: #aee;
   --jp-merge-either-color2: #cff4f4;
-}
\ No newline at end of file
+  --jp-expand-outer-wrapper: rgb(221, 244, 255);
+  --jp-expand-color: rgb(87, 96, 106);
+  --jp-expand-a-color: rgba(84, 174, 255, 0.4)
```
vidartf, line 32: Nitpick: Ideally concepts of the DOM should not bleed into the CSS variable names. Ideally it would say something about the intended difference/relationship between the different variables, so that it can more easily by themed by others (and have those themes survive a change to the DOM).

Also, this color has an alpha element in order to work on multiple themed backgrounds. The other colors in these variables should ideally also work in the same constraints. Mainly since default JupyterLab themes do not include nbdime theming, but we still want nbdime to work with all of them.

gwincr11:
I seem a bit stuck on this since typescript stoped compiling. I opened a pr to update typescript which will allow me to complete this pr, thanks

packages/nbdime/src/styles/variables.css
gwincr11: Ok I changed these to not reflect the dom elements that use them.

packages/nbdime/src/diff/widget/notebook.ts
```diff
@@ -1,45 +1,45 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
-'use strict';
+"use strict";
```
vidartf, line 3: If we rebase this after #667, hopefully these formatting only changes will disappear :)

packages/nbdime/src/diff/widget/notebook.ts
```diff
@@ -1,45 +1,45 @@
...
 /**
  * NotebookDiffWidget
  */
-export
-class NotebookDiffWidget extends Panel {
-  constructor(model: NotebookDiffModel, rendermime: IRenderMimeRegistry) {
+export class NotebookDiffWidget extends Panel {
+  private _model: NotebookDiffModel;
+  private _rendermime: IRenderMimeRegistry;
+  private previousCell: LinkedListCell | null;
```
vidartf, line 32: The previous layout follows the JupyterLab style of having private declarations on the bottom. I think there is value in having consistency in style with lab.

packages/nbdime/src/styles/variables.css
```diff
@@ -27,4 +27,10 @@
 
   --jp-merge-either-color1: #aee;
   --jp-merge-either-color2: #cff4f4;
-}
\ No newline at end of file
+  --jp-expand-outer-wrapper: rgb(221, 244, 255, 1);
+  --jp-expand-color: rgb(87, 96, 106, 1);
+  --jp-expand-activate-color: rgba(84, 174, 255, 0.4);
+  --jp-expand-activate-focus-color: rgb(14, 94, 208, 1);
```
vidartf, line 33: Could these be re-written to be partially transparent, so that they have a fighting chance of working on both dark an light themes? Alternatively be based on JupyterLab CSS variables so that they might have inherent styling?

vidartf:
I opened a rebased version of this PR in #733.

PR #637, opened 2022-10-27
Add a devcontainer

gwincr11:
Add a devcontainer for easy codespace startup. It will start the container with a preconfigured python environment and all needed dev tooling to start workin on the project.

cc @craiglpeters

### DIFF
--- a/.devcontainer/scripts/setup.sh
+++ b/.devcontainer/scripts/setup.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+
+pip install .
+python setup.py build
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,23 @@
 # Contributing
 
 We follow the [IPython Contributing Guide](https://github.com/ipython/ipython/blob/master/CONTRIBUTING.md).
+
+# Contributing with codespaces
+
+If you open this in Codespaces it will pre-install the environment you need to develop with.
+
+## Updating Javascript packages.
+
+This project uses lerna to manage the multiple packages inside the packages folder. If you run npm command at the root it will run the command for each subpackage. This is good for working across packages, however more commonly we want to work with an individual package. To do this change directories into the package you wish to alter, then npm will run only for that package.
+
+### Adding npm dependencie
+
+This should be done inside the package that needs the package installed for.
+
+### Testing changes to the webapp
+
+The python setup.py script will built the entirety of the project and place it inside the build folder. Run `python setup.py build` once you are ready to test a change. Then the nbdime module can be invoked to test changes by running `python -m nbdime service` for example diff web can be started at `python -m nbdime diff-web testnotebookpath testnotebookpath2` there are test notebooks in the `nbdime/webapp/testnotebook` folder.
+
+### Running npm tests
+
+The project uses jest to test the javascript, a typescript compile step will happen prior to testing. You can run the tests for the entire project with `npm run test` in the root, or change directories to the package you are working on and run `npm run test` to test just that package.
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,21 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.245.2/containers/python-3/.devcontainer/base.Dockerfile
+
+# [Choice] Python version (use -bullseye variants on local arm64/Apple Silicon): 3, 3.10, 3.9, 3.8, 3.7, 3.6, 3-bullseye, 3.10-bullseye, 3.9-bullseye, 3.8-bullseye, 3.7-bullseye, 3.6-bullseye, 3-buster, 3.10-buster, 3.9-buster, 3.8-buster, 3.7-buster, 3.6-buster
+ARG VARIANT="3.10-bullseye"
+FROM mcr.microsoft.com/vscode/devcontainers/python:0-${VARIANT}
+
+# [Choice] Node.js version: none, lts/*, 16, 14, 12, 10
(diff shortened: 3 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -1,3 +1,23 @@
...
+
+This should be done inside the package that needs the package installed for.
+
+### Testing changes to the webapp
+
+The python setup.py script will built the entirety of the project and place it inside the build folder. Run `python setup.py build` once you are ready to test a change. Then the nbdime module can be invoked to test changes by running `python -m nbdime service` for example diff web can be started at `python -m nbdime diff-web testnotebookpath testnotebookpath2` there are test notebooks in the `nbdime/webapp/testnotebook` folder.
+
+### Running npm tests
+
+The project uses jest to test the javascript, a typescript compile step will happen prior to testing. You can run the tests for the entire project with `npm run test` in the root, or change directories to the package you are working on and run `npm run test` to test just that package.
```
el-mapache, line 23:
```suggestion
The project uses jest to test the javascript, a typescript compile step occurs before the test suite runs. You can run the tests for the entire project with `npm run test` in the root, or change directories to the package you are working on and run `npm run test` to test just that package.
```

CONTRIBUTING.md
```diff
@@ -1,3 +1,23 @@
 # Contributing
 
 We follow the [IPython Contributing Guide](https://github.com/ipython/ipython/blob/master/CONTRIBUTING.md).
+
+# Contributing with codespaces
+
+If you open this in Codespaces it will pre-install the environment you need to develop with.
```
el-mapache, line 7:
```suggestion
Opening this repository in Codespaces will pre-install the environment you need to develop with.
```

el-mapache (commented):
Thanks for taking this on, love to see a streamlined dev setup! I left a few minor grammatical nits, things that seemed unclear to me as a first-time reader.

.devcontainer/scripts/setup.sh
```diff
@@ -0,0 +1,15 @@
...
+# sudo mkdir /usr/local/share/jupyter
+# sudo chown vscode /usr/local/share/jupyter
+# jupyter labextension develop . --overwrite --no-build
+jupyter server extension enable nbdime # if developing for jupyter lab or nbclassic
+
+jupyter nbextension install --py nbdime --sys-prefix [--sym-link]
+jupyter nbextension enable --py nbdime --sys-prefix
+
+jupyter labextension link ./packages/nbdime --no-build
+jupyter labextension install ./packages/labextension
```
gwincr11, line 15: @vidartf This codespace now includes Jupyter Lab, this will allow people starting up the codespace to have a pre-install Jupyter Lab to develop against. 

However I am not sure the best way to link the local version of nbdime to the JupyterLab server.

This works, but I am not sure how to get changes to show up, I need to rebuild the docker container every time which is not great.

Is there some sort of script I could put together to automatically update Jupyter Lab? Thanks..

gwincr11: I was trying to use this command

`jupyter labextension develop --overwrite .`

But it was not working as I would expect. I wonder if there is some prereq I am not aware of that the cookie cutter does?

gwincr11: Cool thanks. I will dig into this

.devcontainer/scripts/setup.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+pip install .
```
vidartf, line 3: You probably want `-e` flag here to make it an editable install.

scripts/rebuild-lab.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
vidartf, line 1: Generally I'm reluctant to include platform specific scripts in repositories that are meant to be accessible for people across multiple platforms. Is there a reason why these need to be shell scripts, and not node / Python scripts? It makes sense for the .devcontainer since there the OS is controlled by the Dockerfile, but these generic ones are different.

.devcontainer/scripts/setup.sh
vidartf, line 3:
```suggestion
pip install -e .
```

scripts/rebuild-lab.sh
gwincr11: I moved these into the vscode folder and set them up as vscode tasks.

packages/webapp/src/app/diff.ts
```diff
@@ -253,6 +253,7 @@ function initializeDiff() {
     compare(base, remote, 'replace');
   }
 
+  console.log("RUns")
```
vidartf: I assume this was a mistaken inclusion :)

.vscode/scripts/watch-webapp.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
vidartf, line 1: I can't run these shell script tasks on Windows :( If you could write those shell scripts as npm commands, that would be much more platform independent.

.vscode/tasks.json
```diff
@@ -0,0 +1,86 @@
+{
+	"tasks": [
```
vidartf, line 2: In general I would prefer not to include IDE specific files in a repo. Especially since that means that whoever maintains that repo will then have to maintain all the config files for those IDEs, even if they don't use it themselves. Making these tasks into "convenience" npm commands, and documenting them would be a better alternative.

gwincr11:
@vidartf I have made some updates to this to make it more codespaces centric. Wanna kick the tires and see what you think?

gwincr11:
I am just going to simplify this and go for docs for now.

PR #640, opened 2022-11-23
Move global flags to start of regex

singingwolfboy:
This prevents the error "re.error: global flags not at the start of the expression at position 35" on Python 3.11


vidartf (approved):
Thanks!

Eekehgnbm:
Hi, I encountered the error 're.error: global flags not at the start of the expression at position 399' when trying to run 'nengo'. Would this be a similar problem? And how could I adapt the code that was provided by @singingwolfboy to make it fix my error?

PR #647, opened 2022-12-12
add --compatible diff flag to output a diff more compatible with other tools

infokiller:
For context: https://github.com/dandavison/delta/issues/1256

### DIFF
--- a/nbdime/args.py
+++ b/nbdime/args.py
@@ -510,6 +510,13 @@ def add_prettyprint_args(parser):
         default=True,
         help=("prevent use of diff/diff3 for formatting diff/merge text output")
     )
+    parser.add_argument(
+        '--compatible',
+        dest='compatible',
+        action="store_true",
+        default=False,
+        help=("output a diff compatible with other tools")
+    )
 
 
 def prettyprint_config_from_args(arguments, **kwargs):
@@ -520,6 +527,7 @@ def prettyprint_config_from_args(arguments, **kwargs):
         use_color=getattr(arguments, 'use_color', True),
         use_git=getattr(arguments, 'use_git', True),
         use_diff=getattr(arguments, 'use_diff', True),
+        compatible=getattr(arguments, 'compatible', False),
         **kwargs
     )
 
--- a/nbdime/prettyprint.py
+++ b/nbdime/prettyprint.py
@@ -80,7 +80,8 @@ class PrettyPrintConfig:
             use_git = True,
             use_diff = True,
             use_color = True,
-            language = None
+            language = None,
+            compatible = False,
             ):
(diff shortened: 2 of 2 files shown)


infokiller:
@vidartf did you get a chance to look at this? thanks!

vidartf:
I don't really understand what this change is doing. If its trying to output the diff as a proper unified diff, we would need to map the json diffs back to line/character numbers in the original file. There is no trivial way to make this happen. So I assume this change is trying to make some compromise/middle-ground, but not sure exactly what. Please add more details to the PR / command's help key.

infokiller:
> I don't really understand what this change is doing. If its trying to output the diff as a proper unified diff, we would need to map the json diffs back to line/character numbers in the original file. There is no trivial way to make this happen. So I assume this change is trying to make some compromise/middle-ground, but not sure exactly what. Please add more details to the PR / command's help key.

@vidartf sorry for the lack of details, I will try to explain the problem and if you are OK with the solution, I'll also update the CLI help. The problem is described in https://github.com/dandavison/delta/issues/1256 which is an issue I opened because https://github.com/dandavison/delta (syntax highlighting for diffs) didn't work with nbdiff. You can look at the issue for the full details, where the original creator of unified diff also commented there). The TLDR is:

- The output of nbdiff is not compliant with unified diff
- Specifically, the problem with delta is the hunk header was missing (line starting with `@@`)
- I think it's not strictly mandatory to have the correct line/char numbers to be considered "compliant", and in the case of a notebook I guess it may not be easy or even possible
- The unified diff author suggested a different format which avoids `##` lines in order for them to be recognized by unified diff parsers, see https://github.com/dandavison/delta/issues/1256#issuecomment-1345628414

infokiller:
Thanks @vidartf for the quick and helpful response. You suggestions sound good. As for tests, I assume they should go into https://github.com/jupyter/nbdime/tree/master/nbdime/tests is that right?
Are there any specific tests that are good references for what you'd like to see?
Also, should these use as input the higher level ops, or raw notebook?

vidartf:
Yes, I would probably make a new file, and then you could start with something like this:

```python

def test_notebook_diff(any_nb_pair):
    "Test unified diff output on any pair of notebooks in the test suite."
    a, b = any_nb_pair
    diff = diff_notebooks(a, b)

    output = []
    class Printer:
            def write(self, text):
                output.append(text)

    argv = []  # your arguments for diff CLI here
    arguments = _build_arg_parser().parse_args(argv)
    config = prettyprint_config_from_args(args, out=Printer())
    pretty_print_notebook_diff(a.name, b.name, a, diff, config)

    assert "".join(output) == expected_output
```

PR #651, opened 2023-01-29
Replace mock with unittest.mock

yan12125:
The latter is available with Python 3.3 [1], and nbdime requires Python 3.6+ as per setup.py.

This change removes the need of an external library.

[1] https://docs.python.org/3/library/unittest.mock.html

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -50,10 +50,10 @@ jobs:
         python -m pip install jupyterlab~=3.0
         npm install -g codecov
 
-    - name: Use Node.js 12.x
+    - name: Use Node.js 14.x
       uses: actions/setup-node@v1
       with:
-        node-version: 12.x
+        node-version: 14.x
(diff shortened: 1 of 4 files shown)


yan12125:
This pull request is a preparation for adding python-nbdime package to Arch Linux, where it's encouraged to replace mock with unittest.mock: https://archlinux.org/todo/drop-python-mock-checkdepends/

yan12125:
@vidartf I changed some CI settings to fix current errors. Could you re-run failed tests? Also, please tell me if you prefer to split CI changes to a separate pull request.

vidartf:
Thanks for this, and the detailed descriptions in the commit message. If I read this correctly: if we update webpack, we can drop the openssl flag?

yan12125:
> if we update webpack, we can drop the openssl flag?

yep

vidartf:
Merging for now, will look at updating webpack later. Thanks! :)

PR #656, opened 2023-03-24
Add notebook 6 server extension support (w/tests)

vidartf:
Fail loudly (but not catastrophically) if trying to load as a server extension for notebook v6, if jupyter_server 2 is installed. It would fail anyways due to base classes being incompatible, but this will produce a clear error message, and stop loading the nbdime handlers.


fcollonval:
I got hit by the error about `crypto` in jupyterlab-git. I fixed the Jest config like that: https://github.com/jupyterlab/jupyterlab-git/pull/1227/commits/8d0f10a8bce8bc027786ad801d860484151228aa

vidartf:
`codecov` seems to have been pulled from PyPI? Or is PyPI having issues?

PR #668, opened 2023-05-05
Update files with prettier formatting

HaudinFlorence:
Update the .ts, .js, .json, .css and .md files with prettier formatting.


.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
+    "singleQuote": true,
+    "trailingComma": "none",
```
vidartf, line 3: I would prefer trailing commas as often as possible: it leads to smaller diffs in the future, leading to fewer merge conflicts, and a better git blame.

package.json
```diff
@@ -5,6 +5,9 @@
     "build:dev": "tsc --build",
     "clean": "lerna run clean",
     "install": "lerna bootstrap --hoist",
+    "prettier": "prettier --list-different --write \"**/*{.ts,.tsx,.js,.jsx,.css,.json,.md}\"",
+    "prettier:check": "prettier --check \"**/*{.ts,.tsx,.js,.jsx,.css,.json,.md}\"",
+    "prettier:files": "prettier --write",
```
vidartf, line 10: What does the `prettier:files` command do?

vidartf:
Thanks! I left some comments on the config. Additionally it would probably make sense to ignore the package-lock.json file, as it is auto-generated.

.prettierrc
HaudinFlorence: @vidartf I replaced option `"trailingComma": "none"` by `"trailingComma": "all"`, hoping it corresponds to what you require.

package.json
HaudinFlorence: @vidartf To be honest, I took the script I found in JupyterLab's package.json file. I have just removed `prettier:files`.

HaudinFlorence:
@vidartf Sorry for the multiple commit reverts. It is to keep on having all the formatting in a single commit and I forgot to include your suggestion about ignoring the `package-lock.json`.

packages/nbdime/typings/codemirror-metamode.d.ts
```diff
@@ -3,21 +3,19 @@
 // Definitions by: jacqt <https://github.com/vidartf>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-
 // Typings from the codemirror/mode/meta module
```
vidartf, line 6: These typings are pulled from external sources, so probably best to ignore.

packages/nbdime/tslint.json
```diff
@@ -1,58 +1,34 @@
 {
   "rules": {
-    "align": [
-      true,
-      "parameters",
-      "statements"
-    ],
+    "align": [true, "parameters", "statements"],
```
vidartf, line 3: Does the new prettier config still align with tslint config?

.prettierignore
```diff
@@ -0,0 +1,10 @@
+**/build
+**/node_modules
+**/lib
+**/dist
+**/package.json
```
vidartf, line 5: Why are the package.json files ignored?

packages/labextension/style/index.css
```diff
@@ -73,17 +71,17 @@
...
 .nbdime-root.jp-mod-hideUnchanged .jp-Cell-diff.jp-Diff-unchanged {
   display: none;
 }
 
 /* Show a marker with the number of cells hidden before */
-.nbdime-root.jp-mod-hideUnchanged .jp-Cell-diff[data-nbdime-NCellsHiddenBefore]::before,
-.nbdime-root.jp-mod-hideUnchanged .jp-Diff-addremchunk[data-nbdime-NCellsHiddenBefore]::before {
-  content: attr(data-nbdime-NCellsHiddenBefore) " unchanged cell(s) hidden";
+.nbdime-root.jp-mod-hideUnchanged
+  .jp-Cell-diff[data-nbdime-NCellsHiddenBefore]::before,
```
vidartf, line 81: I would prefer it if single CSS rules weren't split across lines (if such a rule config is easily available.

vidartf (commented):
This looks a lot better now! The TS files particularly are ok, but I wonder if the new rules align with what is encoded in the tslint config files Otherwise some small comments and questions.

vidartf:
Note on git techniques: When we agree on the final config, we can use `git rebase -i master`. This should allow for combining the different config commits + formatting commits into one commit each. Then it can be force-pushed with `git push -f`.

packages/labextension/src/plugin.ts
```diff
@@ -1,69 +1,46 @@
 // Copyright (c) Jupyter Development Team.
 // Distributed under the terms of the Modified BSD License.
 
-
 import type {
-  JupyterFrontEndPlugin, JupyterFrontEnd
+  JupyterFrontEndPlugin,
```
bollwyvl, line 5: adding https://github.com/trivago/prettier-plugin-sort-imports , with a bit of config, can help _substantially_ for big import lists e.g. "first unknown-third-party, then lumino, then lab, then local"

.prettierignore
HaudinFlorence: @vidartf I also took the same default config of JupyterLab. I will change that if you prefer.

packages/nbdime/typings/codemirror-metamode.d.ts
HaudinFlorence: @vidartf Ok I can add `typings`directory to the `.prettierignore`file if this is what you mean.

packages/nbdime/tslint.json
HaudinFlorence: As discussed today, you will check this. I haven't done it before to push new commits (interactive rebase).

packages/labextension/style/index.css
HaudinFlorence: I have tried to add `--print-width=120` in prettier script option in the package.json file but it doesn't seem to produce effect on the width of the lines.

packages/nbdime/tslint.json
vidartf: I have checked. There are some small discrepancies, but nothing major, so fine to leave it as it is.

fcollonval:
Thanks @vidartf for pushing on this PR. Could we do anything more to help you merging this one?

vidartf:
No, I think this PR is ready to go, it is just a question of checking the timing vs other PRs in order to minimize conflict resolution work.

krassowski:
@vidartf I wonder if that would have been easier the other way (first merging the lab 4.0 PR and then applying formatting)?

Issue #669, opened 2023-09-05
findAlignedLines sometimes returns unexpected results


HaudinFlorence:
In some case it has been observed that the method `findAlignedLines`used to determine which line numbers align with each other, in the set of DiffViews, return a non correct result, leading to wrong alignment.

We can see here an example of a mergeview:
![Screenshot from 2023-09-05 23-40-52](https://github.com/jupyter/nbdime/assets/99649086/0f9c3a7a-cb5b-4008-8d34-aa6770ac83f0)
The result given by `findAlignedLines`is.
[3, 3, 3, 3]
[4, 5, 4, 5]
[11, 11, 11, 12]
[18, 18, 13, 14]
There is an additionnal padding in the left editor above line 12 causing that this line is not aligned with the 2 lines 11 of central and right editors.

After proposing this manual correction for the array return by `findAlignedLines`, alignment is fixed:
[3, 3, 3, 3]
[4, 5, 4, 5]
[11, **12** 11, 12]
[18, **19**, 13, 14]
![Screenshot from 2023-09-06 00-15-44](https://github.com/jupyter/nbdime/assets/99649086/e647a686-0677-450b-930e-c6ecf6ac961a)

PR #664, opened 2023-04-26
Upgrade nbdime for jupyterlab 4

HaudinFlorence:
This PR is a first pass on trying a migration to jupyterlab4. It includes :
- moving from Lumino 1 to 2
- moving from JupyterLab 2 || 3 to 4 working with codemirror 6 (CM6)

**Main changes**
This PR includes among others the migration to CM6 of these features of the web applications:
- lines and characters hilighting using a dedicated a stateField, stateEffects and specific decorations of CM6
- spacers when addition/deletion of texts are made using a stateField, stateEffects, a builder and decoration widgets of CM6
- picker and conflict gutter markers using dedicated stateFields and stateEffects
- update of the different methods that enable to update the views of the editors and the models of the diffView.

**Diff app rendering example**
comparisons_diff_example8

**Merge app rendering example**
comparisons_merge_example8

**Styling/UX differences**
- the collapse functionnality has been removed.
- the new background is white when it was previously light grey
- in the merge app, the padding spacers don't have colors. They are ll grey (using decoration widgets makes the previous css logics not work anymore, this should be changed to make the coloring possible like in the diff web application.)
- in the diff app, the small arrows between editors have been removed. There are picker gutter markers in the diff too.
- all padding spacers are directly displayed on the view, there is no need to click on a picker gutter markers to make a spacer fully expend with its real sizing.
- there is no picker gutter marker for spacers. In the merge editor, only the text chunks are shown, there is no spacer.
- the syntax highlighting seems to have priority on the character change highlighting (as shown for the `sinus` function in the central editor in the merge example above).
- there is only 1 gutter when there were 2 previously (1 for pickers and 1 for the conflict gutter markers)
- there is a small white space between editors defined in the css of the grid panel now used to encapsulate the editors.
- 'Metadata changed' panel at the end of the applications is missing.

 **Future work and follow up tasks** 
- In some cases, the result of the native function `findAlignedLines` is not right leading to a misalignment of lines to be put at the same height and to incorrect treatment of adding paddings. An issue has been opened with an example #669

- More ui tests will be added when `findAlignedLines` will be fixed for the non working case

- In reference to comment https://github.com/jupyter/nbdime/pull/664#discussion_r1314273776, changes will have to be made after the PR is merged regarding use of  `find`, `each`, `toArray` related to Lumino to move to JS equivalent.

PR #673, opened 2023-09-08
Support JupyterLab 4.0, port to Lumino 2

HaudinFlorence:
This PR is a rebase of https://github.com/jupyter/nbdime/pull/664. It includes fixes following the rebase with the PR adding the linter.
An updated description of the PR is reproduced just below
*************************************************************************************************
PR https://github.com/jupyter/nbdime/pull/664 is a first pass on trying a migration to jupyterlab4. It includes :
- moving from Lumino 1 to 2
- moving from JupyterLab 2 || 3 to 4 working with codemirror 6 (CM6)
- adding a linter
- adding some first ui-tests

**Main changes**
This PR includes among others the migration to CM6 of these features of the web applications:
- lines and characters hilighting using a dedicated a stateField, stateEffects and specific decorations of CM6
- spacers when addition/deletion of texts are made using a stateField, stateEffects, a builder and decoration widgets of CM6
- picker and conflict gutter markers using dedicated stateFields and stateEffects
- update of the different methods that enable to update the views of the editors and the models of the diffView.

**Diff app rendering example**
comparisons_diff_example8

**Merge app rendering example**
comparisons_merge_example8

**Styling/UX differences**
features to be restored
- the collapse functionnality has been removed. It may be re-introduced in a future PR. Not sure how difficult it will be with CM6 (opinions on this point is welcomed).
- the new background is white when it was previously light grey: it disappears in the migration process but it should be rather easy to re-introduce.
- in the merge app, the padding spacers don't have colors. They are all grey (using decoration widgets makes the previous css logics not work anymore, this should be changed to make the coloring possible like in the diff web application.)
 - 'Metadata changed' panel at the end of the applications is missing. This should be restored.

chosen differences or differences related to CM6
- all padding spacers are directly displayed on the view, there is no need to click on a picker gutter markers to make a spacer fully expend with its real sizing.
- there is no picker gutter marker for spacers. In the merge editor, only the text chunks are shown, there is no spacer.
- in the diff app, the small arrows between editors have been removed. There are picker gutter markers in the diff too.
- the syntax highlighting seems to have priority on the character change highlighting (as shown for the `sinus` function in the central editor in the merge example above).
- there is only 1 gutter when there were 2 previously (1 for pickers and 1 for the conflict gutter markers)
- there is a small white space between editors defined in the css of the grid panel now used to encapsulate the editors.

**Additionnal comments on changes**
- adding @types/jest was helping with running the tests
- disabling auto-detection of language server appears to fix the fact that the last Python test was failing. Test was working locally but was simply timing out on CI because search for language servers was taking long time on slow CI file system. 
- [update-integration-tests.yml](https://github.com/jupyterlab/extension-template/blob/e7e4c2bd3467aa3c5a8864e438e897c5f0439098/template/.github/workflows/%7B%25%20if%20test%20%25%7Dupdate-integration-tests.yml%7B%25%20endif%20%25%7D) from the extension template has been added for updating snapshots.It enables to then just comment "please update playwright snapshots" text to trigger that workflow.
- note that references to mathjax have been removed since MathJax4 implementation in JupyterLab 4 now ships all static assets by itself, so there is no need for it. Indeed The import in the server, from jupyter_server_mathjax.app import STATIC_ASSETS_PATH and related code have been removed too.

 **Future work and follow up tasks** 
- In some cases, the result of the native function `findAlignedLines` is not right leading to a misalignment of lines to be put at the same height and to incorrect treatment of adding paddings. An issue has been opened with an example #669

- More ui tests will be added when `findAlignedLines` will be fixed for the non working case

- In reference to comment https://github.com/jupyter/nbdime/pull/664#discussion_r1314273776, changes will have to be made after the PR is merged regarding use of  `find`, `each`, `toArray` related to Lumino to move to JS equivalent.


krassowski:
Hooray, green! Two suggestions:
- Could you update the PR description (copy-pasting the description of https://github.com/jupyter/nbdime/pull/664 would be good enough, but maybe there is something to add)
- Since the commit count went from over 80 to less than 20, there are probably places where the intend of a change is no longer documented by the git history; if would be good to add comments in these places. For example, the reason for disabling auto-detection of language servers.
  - while you are at cleaning the git history I would note that, a commit named "Update alignViews in mergeview.ts to correct padding issues (only the…" appears twice.

HaudinFlorence:
@krassowski Thanks for your suggestion. I have addressed the first point concerning the PR description and will try to take second comment into account.

HaudinFlorence:
> * Since the commit count went from over 80 to less than 20, there are probably places where the intend of a change is no longer documented by the git history; if would be good to add comments in these places. For example, the reason for disabling auto-detection of language servers.

@krassowski Sorry if it is obvious, but where do you think I should document document theses changes?

krassowski:
I would just add inline comments for cases where the internet may not be obvious.

HaudinFlorence:
> I would just add inline comments for cases where the internet may not be obvious.

I have added a section in the description of the PR called "Additionnal comments on changes".

SylvainCorlay (approved):

SylvainCorlay:
I think we should go ahead with this!

JohanMabille (approved):

krassowski (approved):

fcollonval (approved):
Thanks a lot to all of you and especially @HaudinFlorence

SylvainCorlay:
Congrats on this Florence! This unlocks a lot of updates in a large number of JupyterLab extensions!

krassowski:
During the lab call @vidartf asked about dark theme. I can confirm that for JupyterLab diff mode works in dark mode.

| Lab 3.0 | This PR |
|--|--|
| Screenshot from 2023-09-13 20-50-21 | Screenshot from 2023-09-13 20-44-25 |

It would be good to restore the background of cells in dark mode, which in my opinion could be in a follow-up PR. The comments on changed lines are hard to read bu this is not a regression.

nbdime/tests/conftest.py
```diff
@@ -506,6 +506,7 @@ def _kill_nb_app():
          '--port=%i' % port,
         '--ip=127.0.0.1',
         '--log-level=DEBUG',
+        '--LanguageServerManager.autodetect=False', #' adding this fixed the last failing Python test, that was working locally and simply timing out on CI because search for language servers was taking long time on slow CI file system.
```
vidartf, line 509: Would you mind including the name of the test that was failing? This seems like it should be an issue to report upstream, and then we can remove this flag when it is fixed.

nbdime/tests/test_server_extension.py
```diff
@@ -73,15 +73,13 @@ def test_git_difftool(git_repo2, server_extension_app):
...
-    assert data.pop("mathjaxUrl", "").endswith("MathJax.js")
     assert data == {
         "base": "git:",
         "baseUrl": "/nbdime",
         "closable": False,
         "remote": "",
         "savable": False,
-        "hideUnchanged": True,
-        "mathjaxConfig": "TeX-AMS-MML_HTMLorMML-full,Safe",
+        "hideUnchanged": True
```
vidartf, line 82: In Python it is preferable to leave trailing commas in place, as it reduces diff churn.

nbdime/webapp/nbdimeserver.py
```diff
@@ -365,10 +362,7 @@ def make_app(**params):
         (r'/api/diff', ApiDiffHandler, params),
         (r'/api/merge', ApiMergeHandler, params),
         (r'/api/store', ApiMergeStoreHandler, params),
-        (r'/api/closetool', ApiCloseHandler, params),
-        (r'/nb-static/mathjax/(.*)', web.StaticFileHandler, {
-            'path': STATIC_ASSETS_PATH
-        })
+        (r'/api/closetool', ApiCloseHandler, params)
```
vidartf, line 365: comma

package.json
```diff
@@ -15,9 +15,21 @@
     "updated": "lerna updated",
     "watch": "tsc --build --watch"
   },
+  "dependencies": {
+    "npm-which": "^3.0.1"
```
vidartf, line 19: Please put development dependencies as "devDependencies". This helps tools like dependabot determine if the runtime is affected by vulnerabilities or not, or only development tools.

packages/labextension/package.json
```diff
@@ -42,24 +41,25 @@
...
-    "@jupyterlab/nbformat": "^2 || ^3",
-    "@jupyterlab/notebook": "^2 || ^3",
-    "@jupyterlab/rendermime": "^2 || ^3",
-    "@jupyterlab/services": "^5 || ^6",
-    "@jupyterlab/settingregistry": "^2 || ^3",
-    "@lumino/algorithm": "^1.1.2",
-    "@lumino/coreutils": "^1.3.0",
-    "@lumino/disposable": "^1.1.2",
-    "@lumino/widgets": "^1.6.0",
+    "@jupyterlab/application": "^4.0.0",
```
vidartf, line 44: I see you moved some packages from "devDependencies" to "dependencies" here. Was this intentional?

packages/labextension/package.json
```diff
@@ -80,13 +80,14 @@
     },
     "outputDir": "../../nbdime/labextension",
     "sharedPackages": {
-      "codemirror": {
```
vidartf, line 83: I thought this helped reducing the bundle size (as CM would not be needed in the lab bundle). Did I misunderstand? I never did get a full understanding of all the aspects of this setting.

packages/labextension/src/actions.ts
```diff
@@ -2,6 +2,8 @@ import { PathExt, URLExt } from '@jupyterlab/coreutils';
 
 import type { IRenderMimeRegistry } from '@jupyterlab/rendermime';
 
+import { CodeEditor } from '@jupyterlab/codeeditor';
```
vidartf, line 5:
```suggestion
import type { CodeEditor } from '@jupyterlab/codeeditor';
```

packages/labextension/style/index.css
```diff
@@ -1,132 +1,5 @@
...
-  background-color: var(--jp-layout-color2);
-  border-top: solid var(--jp-layout-color3) 1px;
-  border-bottom: solid var(--jp-layout-color3) 1px;
-  text-align: center;
-}
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+@import url('base.css');
```
vidartf, line 5: This file is now empty except for importing `base.css`. Why not simply leave the content of the file here?

packages/nbdime/package.json
```diff
@@ -19,26 +19,32 @@
...
-    "@lumino/coreutils": "^1.3.0",
-    "@lumino/dragdrop": "^1.3.0",
-    "@lumino/signaling": "^1.2.2",
-    "@lumino/widgets": "^1.6.0",
+    "@codemirror/lang-markdown": "^6.1.1",
+    "@codemirror/language": "^6.6.0",
+    "@codemirror/legacy-modes": "^6.3.2",
+    "@codemirror/state": "^6.2.0",
+    "@codemirror/view": "^6.14.0",
+    "@jupyterlab/apputils": "^4.0.0",
```
vidartf, line 27: Also here dev deps became runtime deps. Intentional?

packages/nbdime/package.json
```diff
@@ -50,8 +56,5 @@
     "rimraf": "^5.0.0",
     "ts-jest": "^29.1.0",
     "typescript": "^4.9.0"
-  },
-  "peerDependencies": {
-    "codemirror": "^5.0.0"
```
vidartf, line 55: do wee have peer dependencies on other codemirror packages?

packages/nbdime/src/diff/widget/cell.ts
```diff
@@ -4,12 +4,16 @@
...
 
+import { CodeEditor } from '@jupyterlab/codeeditor';
+
 import { IRenderMimeRegistry, MimeModel } from '@jupyterlab/rendermime';
 
 import { FlexPanel } from '../../upstreaming/flexpanel';
 
 import { CollapsiblePanel } from '../../common/collapsiblepanel';
 
+import { ICellDiffWidgetOptions } from '../../common/interfaces';
```
vidartf, line 15: Why is this not declared in this file? It seems specific to this widget rather than a common interface?

packages/webapp/src/style/index.js
```diff
@@ -0,0 +1,6 @@
+/*
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+import '@jupyterlab/notebook/style/index.js';
+import './base.css';
```
vidartf, line 6: Does this reorg have a purpose in this package (private)?

setup.py
```diff
@@ -128,8 +128,8 @@
     'requests',
     'GitPython!=2.1.4, !=2.1.5, !=2.1.6',  # For difftool taking git refs
     'jupyter_server',
-    'jupyter_server_mathjax>=0.2.2',
     'jinja2>=2.9',
+    # Note that references to jupyter_server_mathjax have been removed since MathJax4 implementation in JupyterLab 4 now ships all static assets by itself, so there is no need for it. Indeed The import in the server, from jupyter_server_mathjax.app import STATIC_ASSETS_PATH.
```
vidartf, line 132: This comment can probably be removed?

tsconfig.json
```diff
@@ -5,5 +5,9 @@
     { "path": "./packages/nbdime" },
     { "path": "./packages/webapp" },
     { "path": "./packages/labextension" }
-  ]
+  ],
+
+  "compilerOptions": {
+    "types": ["node"]
```
vidartf, line 11: Previously, "types" had been left undefined on purpose, because TS would then automatically resolve which types it needed. Is this no longer the case?

tsconfig_base.json
```diff
@@ -16,7 +16,6 @@
     "esModuleInterop": true,
     "noImplicitReturns": true,
     "resolveJsonModule": true,
-    "importsNotUsedAsValues": "error",
```
vidartf, line 19: Can we please keep this?

examples/example1
```diff
@@ -0,0 +1 @@
+../ui-tests/data/merge_test1
```
vidartf, line 1: What are the files in this examples folder meant to be used for? How does it differ from the files in the ui-tests folder, and the nbdime/tests/files folder?

vidartf (commented):
Initial diff scan of the files that do not include editor/mergeview. Doing this first, as it should be quicker to discuss questions for these.

packages/labextension/package.json
HaudinFlorence: Packages will be moved as it was previously.

nbdime/tests/conftest.py
HaudinFlorence: @krassowski I don't remember exactly which python test was failing when you proposed this fix https://github.com/HaudinFlorence/nbdime/pull/7. Could you please remind me here ? Thanks.

packages/labextension/package.json
fcollonval: The shared packages are pre-populated by JupyterLab and some of the core dependencies (through the builder) - this includes the new `@codemirror/` packages as some have requirements to be singleton to work.

So in summary, we can drop codemirror here it is implicitly done for us by JupyterLab.

packages/webapp/src/style/index.js
HaudinFlorence: It will be removed.

nbdime/tests/conftest.py
krassowski: I do not see a path forward to fix this yet. Suggestions welcome. There are no tests timing out elsewhere, but in JupyterLab core I added this line to tests to because there is no benefit in searching the filesystem for language serves when not needed https://github.com/jupyterlab/jupyterlab/pull/15061

```
==================================== ERRORS ====================================
____________________ ERROR at setup of test_isgit[notebook] ____________________

    def _kill_nb_app():
        try:
            process.terminate()
        except OSError:
            # already dead
            pass
>       popen_wait(process, 10)

process    = <Popen: returncode: None args: ['/opt/hostedtoolcache/Python/3.11.4/x64/bin/...>

/opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/site-packages/nbdime/tests/conftest.py:494: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/site-packages/nbdime/tests/conftest.py:29: in popen_wait
    return p.wait(timeout)
        p          = <Popen: returncode: None args: ['/opt/hostedtoolcache/Python/3.11.4/x64/bin/...>
        timeout    = 10
/opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/subprocess.py:1264: in wait
    return self._wait(timeout=timeout)
        endtime    = 256.32528873
        self       = <Popen: returncode: None args: ['/opt/hostedtoolcache/Python/3.11.4/x64/bin/...>
        timeout    = 10
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <Popen: returncode: None args: ['/opt/hostedtoolcache/Python/3.11.4/x64/bin/...>
timeout = 10

    def _wait(self, timeout):
        """Internal implementation of wait() on POSIX."""
        if self.returncode is not None:
            return self.returncode
    
        if timeout is not None:
            endtime = _time() + timeout
            # Enter a busy loop if we have a timeout.  This busy loop was
            # cribbed from Lib/threading.py in Thread.wait() at r71065.
            delay = 0.0005 # 500 us -> initial delay of 1 ms
            while True:
                if self._waitpid_lock.acquire(False):
                    try:
                        if self.returncode is not None:
                            break  # Another thread waited.
                        (pid, sts) = self._try_wait(os.WNOHANG)
                        assert pid == self.pid or pid == 0
                        if pid == self.pid:
                            self._handle_exitstatus(sts)
                            break
                    finally:
                        self._waitpid_lock.release()
                remaining = self._remaining_time(endtime)
                if remaining <= 0:
>                   raise TimeoutExpired(self.args, timeout)
E                   subprocess.TimeoutExpired: Command '['/opt/hostedtoolcache/Python/3.11.4/x64/bin/python', '-m', 'jupyter_server', '--port=62026', '--ip=127.0.0.1', '--log-level=DEBUG', '--no-browser', '--IdentityProvider.token=nbdime-test-token']' timed out after 10 seconds

delay      = 0.049[130](https://github.com/krassowski/nbdime/actions/runs/6045434096/job/16405481322#step:6:131)42399994083
endtime    = 256.32529052999996
pid        = 0
remaining  = -0.000198[166](https://github.com/krassowski/nbdime/actions/runs/6045434096/job/16405481322#step:6:167)0000[183](https://github.com/krassowski/nbdime/actions/runs/6045434096/job/16405481322#step:6:184)183
self       = <Popen: returncode: None args: ['/opt/hostedtoolcache/Python/3.11.4/x64/bin/...>
sts        = 0
timeout    = 10

/opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/subprocess.py:2038: TimeoutExpired
```

vidartf:
Re the OP's "Styling/UX differences" sections. Is that up to date? Would you mind clarifying which of these points are changed because of technical challenges, and which (if any) are considered improvements? E.g. why was the ability to have collapsed sections removed, and what would the challenges be to add it back in?

packages/nbdime/package.json
HaudinFlorence: This will be removed as it was previously for the common dependencies.

packages/nbdime/src/diff/widget/cell.ts
HaudinFlorence: It is also used here: `packages/nbdime/src/merge/widget/cell.ts`

nbdime/tests/conftest.py
vidartf: Ok, let's just leave this one as it is then, but it seems weird to me that the LSP autodetect is a blocking operation...

tsconfig_base.json
HaudinFlorence: It will be re-established.

HaudinFlorence:
> Re the OP's "Styling/UX differences" sections. Is that up to date? Would you mind clarifying which of these points are changed because of technical challenges, and which (if any) are considered improvements? E.g. why was the ability to have collapsed sections removed, and what would the challenges be to add it back in?

Thanks @vidartf for your review. Concerning the styling/UX differences sections, it is up to date and I will clarify why these points are changed.

package.json
HaudinFlorence: Dependencies should have been restored as they were.

setup.py
HaudinFlorence: It will be removed

tsconfig.json
HaudinFlorence: `compilerOptions` will be removed.

packages/nbdime/package.json
HaudinFlorence: It seems that we don't need peer dependencies.

nbdime/webapp/nbdimeserver.py
```diff
@@ -50,9 +49,7 @@ def base_args(self):
             'closable': self.params.get('closable', False),
             'savable': fn is not None,
             'baseUrl': self.nbdime_base_url,
-            'hideUnchanged': self.params.get('hide_unchanged', True),
-            'mathjaxUrl': self.mathjax_url,
-            'mathjaxConfig': self.mathjax_config,
+            'hideUnchanged': self.params.get('hide_unchanged', True)
```
vidartf, line 52: another trailing comma (we could fix these with `black` later, but now good for diff)

packages/labextension/package.json
vidartf: Thanks for the update 👍

packages/labextension/style/index.css
```diff
@@ -1,132 +1,145 @@
...
+ @import url('nbdime/lib/common/dragpanel.css');
+ @import url('nbdime/lib/styles/variables.css');
+ @import url('nbdime/lib/styles/common.css');
+ @import url('nbdime/lib/styles/diff.css');
+ @import url('nbdime/lib/styles/merge.css');
+
+ .nbdime-Widget {
+   display: flex;
+   flex-direction: column;
+ }
```
vidartf, line 16: For me, this file renders as a full diff, even though there is very little changes. Is there a change to line endings or something? 😕

vidartf: Actually, I think there is an addition of 1 space indent in this file now?

packages/nbdime/src/diff/widget/metadata.ts
```diff
@@ -18,8 +22,13 @@ const ROOT_METADATA_CLASS = 'jp-Metadata-diff';
  * MetadataWidget for changes to Notebook-level metadata
  */
 export class MetadataDiffWidget extends Panel {
-  constructor(model: IStringDiffModel) {
+  // TODO improve typing hierarchy to avoid `Omit`
```
vidartf, line 25: Is this TODO for yourself, or for someone in the future? :)

examples/example1
HaudinFlorence: Maybe we could reduce the number of examples in the dedicated folders.They were used during the migration process to check various diff and merge configurations. 
The diff in `ui-tests/diff_test1` and  `ui-tests/diff_test2` are variants of example1 (center.ipynb with addition/deletions of additionnal cells) and `ui-tests/diff_test3` is a variant of `example8` (with `left.ipynb` and `center.ipynb` notebooks). 
The merge in `ui-tests/merge_test1` and `ui-tests/merge_test2` are variant of example1 and  `ui-tests/merge_test3` is a variant of `example3`. 

`nbdime/tests/files folder` has nothing in common with `ui-tests` and `examples`.

packages/nbdime/src/diff/widget/metadata.ts
HaudinFlorence: I don't think this comment is mine.

packages/labextension/style/index.css
HaudinFlorence: @vidartf Both the comma and the indentation issues have been fixed in commit

HaudinFlorence: > Actually, I think there is an addition of 1 space indent in this file now?

Yes there was. Introduced with a copy-paste and I did not see indentation was changed.

packages/nbdime/src/diff/widget/metadata.ts
fcollonval: I added it - I will remove it when working into the integration with `@jupyterlab/git`

fcollonval:
@vidartf are you concerns addressed? Could we merge this PR?

Could you also accept the NPM invitation to join the jupyter organization? Once accepted I will have to increase your rights for you to add the nbdime package as it is not possible to grant new member advanced rights by default.

vidartf:
@fcollonval My previous first-pass concerns are indeed addressed, and I've convinced myself that the styling concerns I have can be iterated on in follow-up PRs (I have fixes for most of them locally). My current major concern is that the mergetool seems to be extremely unreliable (clicking the pickers sometimes work, and sometimes not at all), and I am trying to figure out exactly what is causing this. I'm going to spend some time today to understand it, and how much work a fix would entail, then I'll get back to you.

vidartf:
Never mind the mergetool. The examples I found that broke it seems to be broken in previous releases of nbdime as well. I will create an issue for that separately.

vidartf:
(actually, the mergetool is more generally broken in this PR than previous releases, but I think we can merge and iterate on that...)

SylvainCorlay:
Congrats to @HaudinFlorence, this is a significant achievement!

Many thanks to everyone involved (@vidartf @krassowski @fcollonval @JohanMabille).

vidartf:
@HaudinFlorence I'm not sure if I fully understood what this meant. Can you help clarify?

> Metadata changed' panel at the end of the applications is missing. This should be restored.

HaudinFlorence:
> @HaudinFlorence I'm not sure if I fully understood what this meant. Can you help clarify?
> 
> > Metadata changed' panel at the end of the applications is missing. This should be restored.

It seemed to me that in the case of the merge app (mostly when there are more than one cells), the blue indication after the cell mentioning "Metadada changed" was missing. I am adding a capture of how it looks in CM5 version (not able to build CM6 version right now.)

Screenshot from 2023-09-18 15-04-40

PR #687, opened 2023-09-21
Use pseudo-elements instead of border to preserve line height

### LINKED ISSUE #674: CM6: Fix border styles / line spacing
With CM 6, the chunk highlighting is now with `border` style directly on the line elements. This occupies vertical space. We should consider whether a different styling method can be used, or compensate for this space by different CSS rules.

krassowski:
Fixes #674

This PR re-implements chunk borders with pseudo-elements instead of `border-top`/`border-bottom`.

> Note: The obvious alternative solution of using negative margin does not work because CM6 calculates line height to programmatically set the height of gutter (e.g. line numbers), which does not account for margins.

| Before | After |
|--|--|
| image | image |


fcollonval (approved):
Thanks @krassowski

fcollonval:

fcollonval:
Merging to reduce multiple rebase / update snapshots

Issue #676, opened 2023-09-18
CM6: Fix missing mergeview pickers


vidartf:
After the CM6 upgrade, there is now missing picker handles in a few cases, especially for merge spacers. These should be re-added.

vidartf:
Note, on #673, this was listed as:

> chosen differences or differences related to CM6
    - all padding spacers are directly displayed on the view, there is no need to click on a picker gutter markers to make a spacer fully expend with its real sizing.
    - there is no picker gutter marker for spacers. In the merge editor, only the text chunks are shown, there is no spacer.

I don't really understand this, as the pickers are meant to be used to pick which version of the code to use (local, remote or base).

krassowski:
Somewhat related issue is that clicking on the gutter in a place which does not have a picker leads to an uncaught error:

```js
Uncaught TypeError: o is undefined
    onGutterClick mergeview.ts:1415
    mousedown mergeview.ts:1110
    ar index.js:9652
```

HaudinFlorence:
> Note, on #673, this was listed as:
> 
> > chosen differences or differences related to CM6
> > - all padding spacers are directly displayed on the view, there is no need to click on a picker gutter markers to make a spacer fully expend with its real sizing.
> > - there is no picker gutter marker for spacers. In the merge editor, only the text chunks are shown, there is no spacer.
> 
> I don't really understand this, as the pickers are meant to be used to pick which version of the code to use (local, remote or base).

On the contrary, I don't understand why one needs to click on a picker to make the spacers appear. Maybe I missed something in the logics choosen for nbdime. I thought that the spacers are there to align the different views and are not related to a choice of what code will be picked.

vidartf:
@HaudinFlorence My comment was mainly about the second point "there is no picker gutter marker for spacers. In the merge editor, only the text chunks are shown, there is no spacer". My understanding is that the pickers are needed also when there is added code vs an "empty spacer", so that you can pick the version of the code without the addition. For the first point, I'm not sure what it refers to, but that was at least never the intended behavior, and would sound like an indicator of a bug.

krassowski:
> My understanding is that the pickers are needed also when there is added code vs an "empty spacer", so that you can pick the version of the code without the addition.

So that should be solved by #691 :)

HaudinFlorence:
[...]

PR #688, opened 2023-09-21
Fix wrong picker marker (a typo during lab 4.0 migration)

krassowski:
This is a follow-up to #673, something I noticed when looking into https://github.com/jupyter/nbdime/issues/676

For reference, this is the corresponding code on 3.2 branch:

https://github.com/jupyter/nbdime/blob/1b368a884405914e9b1861a445c20927f7c88a5b/packages/nbdime/src/common/mergeview.ts#L555-L572

it was also using `conflict` so I assume `picker` here on the master branch is a typo.

### DIFF
--- a/packages/nbdime/src/common/mergeview.ts
+++ b/packages/nbdime/src/common/mergeview.ts
@@ -691,7 +691,7 @@ Add a gap DOM element between 2 editors
                 !hasEntries(s.decision.localDiff) &&
                 !hasEntries(s.decision.remoteDiff)
               ) {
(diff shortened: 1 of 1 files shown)


HaudinFlorence:
@krassowski Thanks for pointing this out. This seems indeed to be an error or a copy paste reminiscence.

HaudinFlorence (approved):
@krassowski Thanks for pointing this wrong picker marker. This seems to be an error or a copy paste reminiscence.

HaudinFlorence (approved):

vidartf (approved):

Issue #678, opened 2023-09-18
CM6: Investigate syntax highlighting in diff chunks


vidartf:
From #673:
> the syntax highlighting seems to have priority on the character change highlighting (as shown for the sinus function in the central editor in the merge example above).

krassowski:
Maybe I do not understand the issue but it seems to be an improvement to me, isn't it? As per the screenshot below, in previous version the syntax highlight was lost in the central editor, but in the version updated for JupyterLab/CM6 the syntax highlighting is present, which makes it easier to read the removed code:

![image](https://github.com/jupyter/nbdime/assets/5832902/f504d8d2-c707-4331-b821-7ee47168b032)

It also appears to be in "chosen differences or differences related to CM6" rather than "features to be restored" section so I think it was just being highlighted as an improvement? Or am I missing something and it is all backwards? @HaudinFlorence?

krassowski:
Ok with better zoom and color correction I see that some tokens which are not syntax-highlighted are in dark red:

| Before | After |
|--|--|
|![image](https://github.com/jupyter/nbdime/assets/5832902/ffce4ea8-f089-4a3e-8c03-42e467aca106) | ![image](https://github.com/jupyter/nbdime/assets/5832902/7576600f-25ca-457d-a159-b896e67a11af) |

which is still an improvement I think.

vidartf:
I think the argument is that it should either be fully diff colored (red/green tint) or fully syntax colored (only relying on  background color for add/remove context). The current mix where some parts are red/green tinted, while others are highlighted by syntax is confusing.

HaudinFlorence:
@krassowski @vidartf I agree with the last comment in Vidar. I am proposing this PR to hopefully remove the diff coloring in the mergeview case with 4 editors and choose the full syntax coloring approach.

vidartf:
Closed in  #689

PR #689, opened 2023-09-22
Fix syntax highlighting issue in diffchunks.

HaudinFlorence:
Fixes mixed approach observed in issue #678, where both syntax highlighting and diff chunk text coloring was observed in the central editor of the mergeview.


fcollonval (approved):
Thanks @HaudinFlorence

krassowski (approved):
Thank you!

PR #691, opened 2023-09-23
Fix pickers for empty chunks

### LINKED ISSUE #676: CM6: Fix missing mergeview pickers

krassowski:
Fixes #676.

Currently empty chunks are represented using `PaddingWidget` which is a block widget; block widgets do not have a corresponding line, therefore one cannot place a gutter marker over them using `markers` method; instead a dedicated `widgetMarker` method needs to be defined.

Preview

empty-picker-restored


krassowski:

krassowski:
Kicking CI

fcollonval (approved):
Thanks @krassowski

PR #697, opened 2023-09-27
Add Node.js to docs runtime

krassowski:
This should resolve current failures of documentation build (https://github.com/jupyter/nbdime/issues/685#issuecomment-1735985166).

PR #698, opened 2023-09-28
RTD: make sure nodes is present

minrk:
and build with Python 3.11

builds are [failing](https://readthedocs.org/projects/nbdime/builds/22047074/) with missing `npm`.


fcollonval:
Superseeds #697

PR #706, opened 2023-10-11
Improve `findAlignedLines`

### LINKED ISSUE #669: findAlignedLines sometimes returns unexpected results

fcollonval:
Fixes #669

Compute correctly the alignment for a chunk in another diff view.

**Before**

image

**After**

image


fcollonval:

HaudinFlorence:
@fcollonval thanks for this PR. I have tried it on a case that was not producing correct results (the case reported in issue https://github.com/jupyter/nbdime/issues/669 that is close to the one present in this PR) and it seems to work.

packages/nbdime/src/common/mergeview.ts
```diff
@@ -1619,53 +1629,56 @@ export class MergeView extends Panel {
...
       ? this._base.cm.defaultLineHeight
       : this._diffViews[0].remoteEditorWidget.cm.defaultLineHeight;
     if (this._aligning) {
       return;
     }
     this._aligning = true;
     // Find matching lines
-    let linesToAlign = findAlignedLines(this._diffViews);
+    const linesToAlign = findAlignedLines(this._diffViews);
+    console.log(linesToAlign)
```
vidartf, line 1641: debug log statement left behind

packages/nbdime/src/common/mergeview.ts
```diff
@@ -1179,17 +1179,26 @@ function getMatchingEditLine(baseLine: number, chunks: Chunk[]): number {
...
     }
     if (chunk.baseFrom > editLine) {
+      // Remaining chunks are after the one we are interested
+      previous = chunks[i - 1];
       break;
     }
   }
-  return toMatch.baseTo;
+  // toMatch is not in chunks list, add lines delta from the last chunk
+  console.log(previous, toMatch.baseTo + (previous ? (previous.remoteTo - previous.baseTo) : 0))
```
vidartf, line 1200: debug log statement left behind

packages/nbdime/src/common/mergeview.ts
```diff
@@ -1179,17 +1179,26 @@ function getMatchingEditLine(baseLine: number, chunks: Chunk[]): number {
  *
  */
 function getMatchingEditLineLC(toMatch: Chunk, chunks: Chunk[]): number {
-  let editLine = toMatch.baseFrom;
+  console.log(toMatch, chunks)
```
vidartf, line 1182: log statement

PR #694, opened 2023-09-25
Deal with chunk to align outside of viewport

fcollonval:
Fixes #693

If some alignments are outside of the viewport, this adds an initial spacers to have persist the alignments.

- [ ] Fix test to remove collapsing identical lines...

PR #707, opened 2023-10-11
Make collapseIdentical margin configurable

fcollonval:
Make collapseIdentical margin configurable
Improve base interfaces

This is extracted from #694 that is stuck on windows case.


fcollonval:
Merging to carry out a new pre-release that will allow for a pre-release of jupyterlab-git

PR #708, opened 2023-10-11
Upgrade devdependencies and clean up the configuration

fcollonval:


package.json
```diff
@@ -15,20 +15,13 @@
     "updated": "lerna updated",
     "watch:webapp": "run-p watch:lib watch:app",
     "watch:app": "lerna exec --stream --scope \"nbdime-webapp\" npm run watch",
-    "watch:lib": "lerna exec --stream --scope \"nbdime\" --scope \"jupyterlab-rise\" npm run watch"
+    "watch:lib": "lerna exec --stream --scope \"nbdime\" --scope \"nbdime-jupyterlab\" npm run watch"
   },
   "devDependencies": {
-    "@jupyterlab/buildutils": "4.0.0",
```
vidartf, line 21: Where does `update-dependency` in "update:*" scripts come from now? I cannot run these commands locally anymore after this dep is removed.

PR #710, opened 2023-10-12
Set up releaser

fcollonval:


fcollonval:
For now this is blocked by:
- A bug on the releaser: fix opened upstream https://github.com/jupyter-server/jupyter_releaser/pull/531
- A bug on the CI preventing the bump of NPM package. The command `npx lerna version --no-push --no-git-tag-version -y --force-publish patch` seems to not be running on the CI; see for example https://github.com/jupyter/nbdime/actions/runs/6524031965/job/17715287136?pr=710

fcollonval:
Kicking the CI

PR #712, opened 2023-10-16
Bump version

fcollonval:
Bump version to prepare a new alpha release


vidartf:
PS: You would normally also need to bump the deps from the webapp on nbdime

Issue #675, opened 2023-09-18
CM6: Fix background color


vidartf:
In lab 4, the theme for a code editor is only scoped on the notebook and console classes (in a bit of a bleedthrough in the lab codebase: the specific extensions "notebook" and "console" are special cased in the generic "codemirror" package): 

https://github.com/jupyterlab/jupyterlab/blob/1d989bad87e89307a336d2f671c9cc98ec37f5e9/packages/codemirror/src/theme.ts#L27C1-L29

We should either see if we can add to the global theme such that our diff/merge views also get this styling, or create a specific selector outside of the theme system.

HaudinFlorence:
> We should either see if we can add to the global theme such that our diff/merge views also get this styling, or create a specific selector outside of the theme system.

@vidartf I can try to work on it. Is there a special reason to have a grey background and not something closer to notebook cells ? What is best to choose between the 2 options mentioned above.

vidartf:
> grey background and not something closer to notebook cells ? 

To clarify: Notebook cells *do* have a gray background. Any usage of codemirror outside of notebook and console on the other hand now (with lab 4) has a default background of white. We should try to mimic the look in notebooks as much as possible, so we should also use the same theming.

vidartf:
> What is best to choose between the 2 options mentioned above.

The first option is probably more robust / future-proof, but the second option will be simpler to implement (the way CM styling works now).

HaudinFlorence:
@vidartf Thanks for your input and for the precisions. I opened a PR to try to fix the issue https://github.com/jupyter/nbdime/pull/714.

HaudinFlorence:
I think that this issue can be closed.

vidartf:
Thanks

PR #714, opened 2023-10-16
Fix editor background color

HaudinFlorence:
This PR aims at fixing the missing background color for the codemirror editors, to be aligned with jupyterlab cells' color (when unselected). It adresses the case of the webapp. The background is not yet fixed for the labextension.

The re-introduced background color is the one associated with the css variable `--jp-cell-editor-background` as can be seen on the following capture:

Screenshot from 2023-10-16 17-46-55


fcollonval (approved):
Thanks @HaudinFlorence

krassowski:
I have two questions here:

> The background is not yet fixed for the labextension.

1. In that case should we replace the "fixes" keyword so that the issue https://github.com/jupyter/nbdime/issues/675 does not get closed on merging this PR?

2. Why are there no changes to the snapshots? Is it not covered by snapshots, or is it that the snapshots do differ but the change in colour is below sensitivity threshold and did not get picked up? If it is the latter, shall we tighten the threshold to ensure we catch visual regressions in the bg color?

HaudinFlorence:
bot please update snapshots

HaudinFlorence:
@krassowski 

> I have two questions here:
> 
> > The background is not yet fixed for the labextension.
> 
> 1. In that case should we replace the "fixes" keyword so that the issue [CM6: Fix background color #675](https://github.com/jupyter/nbdime/issues/675) does not get closed on merging this PR?

In fact, locally the change was not visible in my labextension. Could you please check on your local one? Normally the backgorund should be taken into account because the changes are introduced in `nbdime/lib/styles/common.css` that is imported in `labextension/style/index.css`

> 2. Why are there no changes to the snapshots? Is it not covered by snapshots, or is it that the snapshots do differ but the change in colour is below sensitivity threshold and did not get picked up? If it is the latter, shall we tighten the threshold to ensure we catch visual regressions in the bg color?

 I think it was only a matter on updates.

krassowski:
> I think it was only a matter on updating the snapshots.

Well, if the tests were all green before updating the snapshots it strongly suggests that we should adjust `maxDiffPixelRatio` or `maxDiffPixels` to catch a potential regression in the future.

Reference: https://playwright.dev/docs/api/class-pageassertions#page-assertions-to-have-screenshot-1

HaudinFlorence:
bot please update snapshots

HaudinFlorence:
> > I think it was only a matter on updating the snapshots.
> 
> Well, if the tests were all green before updating the snapshots it strongly suggests that we should adjust `maxDiffPixelRatio` or `maxDiffPixels` to catch a potential regression in the future.
> 
> Reference: https://playwright.dev/docs/api/class-pageassertions#page-assertions-to-have-screenshot-1

I am not completely sure that all tests were green. I take note about your suggestion.

krassowski:
The windows snapshots seem to have been inadvertly deleted.

ui-tests/tests/nbdime-merge-test6.spec.ts-snapshots/merge-test6-take-a-snapshot-at-opening-1-linux.png
krassowski, line 1: There appears to be a regression with the gutter for the "cell added remotely". Is it related to this PR?

fcollonval: I have seen wrong number positioning transiently. I triggered the tests to see if this particular test is passing or failing.

krassowski:
Good call, failed in tests.

krassowski (approved):
Thank you @HaudinFlorence, snapshots look good now.

PR #728, opened 2023-10-31
CM6: Fix spacing between editors

HaudinFlorence:
This PR fixes #https://github.com/jupyter/nbdime/issues/720

The gap between the different editors that was put to 10 px has been removed.

**CM5 version**
Screenshot from 2023-10-31 18-23-51

**CM6 version**
Screenshot from 2023-10-31 18-27-02

Issue #729, opened 2023-11-01
Gutter pickers are too small if "DejaVu Sans Mono" font is used.


vidartf:
Running UI-tests locally, I always get a diff showing my local version having much smaller gutter pickers/conflict markers for a merge. I see locally that I end up using the "DejaVu Sans Mono" font via the following computed font-family: `menlo, consolas, "DejaVu Sans Mono", monospace sans-serif`. 

This is how it looks:
![image](https://github.com/jupyter/nbdime/assets/510760/42ed213b-3c6f-4e3b-81e3-7cb1d26f6a8b)

We should probably override the font-family setting for the mergeview gutters, such that it avoids this font face? I think the main issue is that I have neither menlo or consolas locally?

Note: This issue precedes the CM6 work, but the UI-tests helped highlight the issue during local dev.

PR #731, opened 2023-11-06
Ensure collapsed mergepane border

### LINKED ISSUE #732: CM6: Conflicting cell ops (deleted/added) are missing color highlighting
Old look (CM5):
image

New look (CM6):
image

While it is still possible to distinguish which cells as being added or removed, we should ensure there is still color highlighting via background color to these full cells to make it easier to visually distinguish.

vidartf:
Followup to #728  After switching to grid based layout, each pane had its own border. With 0px gap, the border still did not collapse to single 1px border, so instead we switch to using gap + background-color.

C.f. https://stackoverflow.com/questions/47882924/preventing-double-borders-in-css-grid

Fixes #732.


vidartf:

vidartf:

ui-tests/tests/nbdime-diff-test1.spec.ts-snapshots/diff-test1-take-a-snapshot-at-opening-1-linux.png
krassowski, line 1: It feels like the old version (a single green border) looked cleaner than the new one (a green border in grey outline).

ui-tests/tests/nbdime-diff-test3.spec.ts-snapshots/diff-test3-take-a-snapshot-at-opening-1-linux.png
krassowski, line 1: It is nicer to here indeed.

ui-tests/tests/nbdime-diff-test1.spec.ts-snapshots/diff-test1-take-a-snapshot-at-opening-1-linux.png
vidartf: I think you are right, probably the rule shouldn't be added for the 1-way.

vidartf:

vidartf:
Attempted tweak in++, and now also fixes #732.

packages/nbdime/src/styles/common.css
```diff
@@ -1,3 +1,9 @@
+
```
krassowski, line 1:
```suggestion
```

krassowski (approved):
It looks that instead of collapsing the border it now adds a tiny gap which also appears fine I think.

vidartf:
> It looks that instead of collapsing the border it now adds a tiny gap which also appears fine I think.

Thanks for pointing it out. Rerending via the bot was the last thing I did on Wednesday EOD. I rely on it since #729 prevents UI test from being useful locally for me.

vidartf:

PR #735, opened 2023-11-10
Fix duplicate pickers

### LINKED ISSUE #700: CM6: Several conflict gutter markers placed on the same line
When trying to run example6, the following issue has been observed in the merge editor: several conflict gutter markers are produced at the same line (case of line 6), not placed exactly at the same location.

Screenshot from 2023-10-02 15-24-57

vidartf:
To avoid remote, local and merge diff views adding overlapping gutter markers of the same type on the same line. We leave it to CM to handle multiple markers of different types on the same line, but that shouldn't really apply for us.

Fixes #700 .


vidartf:
(to have a cleaner diff, exclude the lint commit from the range)

vidartf:
Hmm... The UI tests shouldn't be passing without a rerender, since the bug is in one of the UI tests, so something went wrong here..

vidartf:

packages/nbdime/src/common/mergeview.ts
```diff
@@ -295,21 +295,21 @@ const removeGutterMarkerEffect = StateEffect.define<{ type: string }>({
...
-    pickerMarker.className = this._className;
+    let pickerMarker = elt('div', this.symbol);
+    pickerMarker.className = this.className;
     return pickerMarker;
   }
-  private _symbol: string;
-  private _className: string;
-  private _block: boolean;
+  eq(other: GutterMarker): boolean {
+    return other === this;
```
krassowski, line 308: Should it check for equality of symbol/className/block instead?

vidartf: I wondered about it, but we are using static instances of the markers here, so I figured it would be simpler with an identity check instead. I was actually surprised I had to write this, as the [docs for `RangeValue`](https://codemirror.net/docs/ref/#state.RangeValue.eq) says:

> [...] Used when comparing rangesets. The default implementation compares by identity. [...]

`GutterMarker` extends `RangeValue`, and has its own docs for `eq` which doesn't say anything about its defaults, but in the code it always returns false:

https://github.com/codemirror/view/blob/cf5e5a9e2f349a3ca421107e0502028842ea526b/src/gutter.ts#L19

vidartf: That code also has some other stuff like an internal `compare`, and a function `sameMarkers`, that both do a bit more...

krassowski:

krassowski (approved):
Thanks!

PR #736, opened 2023-11-17
Change header for unchanged notebook meta

### LINKED ISSUE #718: Identical content editor are collapsed
From jupyterlab-git example, single cell added or unchanged have their content collapse as identical. We should not collapse the content in such case.

image

There is also the case of the notebook metadata. But the issue is larger as the collapse widget title is wrong by announcing there is a change even when there is none. And in addition, it displays 4 editors even if there are no changes.

vidartf:
We previously changed to always showing the metadata diff in notebook merges, in case the user needs to make manual edits to it even if there are no changes in the merge itself. That change did not take into account that the display behavior here made some assumptions about only being shown when there was a change. Here we make the title conditional on whether there are changes or not, and make sure not to collapse similar changes in the case where there are none.

Fixes #718 .


vidartf:

ui-tests/tests/nbdime-merge-test1.spec.ts-snapshots/merge-test1-should-not-collapse-source-for-unchanged-metadata-1-linux.png
vidartf, line 1: What is wrong with the size of the top editors here? Part of the document is missing, and the line number scrolling is off...

vidartf: Ah, probably just the CSS transition effect, I'll add a `sleep` or similar.

vidartf:

packages/nbdime/src/merge/widget/metadata.ts
```diff
@@ -36,18 +36,27 @@ export class MetadataMergeWidget extends MergePanel<MetadataMergeModel> {
...
     });
     const trans = this._translator.load('nbdime');
     const wrapper = new CollapsiblePanel(
       this.view,
-      trans.__('Notebook metadata changed'),
+      trans.__(
+        `Notebook metadata ${
+          unchanged ? 'unchanged' : 'changed'
+        }`,
+      ),
```
krassowski, line 59: This needs to be two separate strings or use %1 with separate `trans.__()` wrappers for each unchanged/changed words. It is slightly easier for translators if they have two full sentences rather than three strings they need to understand how they match together.

```suggestion
      unchanged ? trans.__('Notebook metadata unchanged') : trans.__('Notebook metadata changed'),
```

krassowski (approved):
Thank you!

Issue #659, opened 2023-04-14
Migration to jupyterlab 4.0


HaudinFlorence:
Hello,
@vidartf I am about to work on the migration of jupyterlab-git extension to jupyterlab 4.0 that depends on nbdime. Could you please give  a hint about what will be the strategy concerning nbdime: is it planed to keep the compatibility with previous versions of jupyterlab or not?
Many thanks in advance.
cc @fcollonval 

```[tasklist]
### Tasks
- [ ] https://github.com/jupyter/nbdime/issues/674
- [ ] https://github.com/jupyter/nbdime/issues/675
- [ ] https://github.com/jupyter/nbdime/issues/676
- [ ] https://github.com/jupyter/nbdime/issues/679
- [ ] https://github.com/jupyter/nbdime/issues/700
- [ ] https://github.com/jupyter/nbdime/issues/721
- [ ] https://github.com/jupyter/nbdime/issues/720
- [ ] https://github.com/jupyter/nbdime/pull/639
```

vidartf:
Hi,
I'm not sure yet how much work will be needed to add support for lab 4.0. I believe the main things of importance are:
- any changes for lumino 2.0. If there are any breaking changes that are not easy to support both at the same time, it can be hard to support both 4.0 and 3.0 with the same version. However, I think the actual change is for nbdime to support 3 + 4 instead of the current 2 + 3.
- any changes related to the codemirror version bump. I don' think it should be an issue, but maybe there will be some conflict if we try to put both on the page at the same time?
- others??

vidartf:
@fcollonval To try to rephrase myself: 

I want to add support for JupyterLab 4. The amount of work needed to do that depends on how many things we use break with the lumino 2.0 upgrade. If there are no breakages, we might even be able to support both 3.0 and 4.0.

Separately there is the question of the Codemirror change in JupyterLab 4.0. As of yet, the CM "DiffView/MergeView" code that I used as a basis for https://github.com/jupyter/nbdime/blob/master/packages/nbdime/src/common/mergeview.ts has not been ported to CM6, so I do not think that we can quickly change to using CM6 for nbdime. Hopefully, it should not be a problem for us to continue using CM5 in nbdime with JupyterLab 4.0 though. I'd be interested to hear your thoughts on whether both of them can be on the page at the same time.

vidartf:
Or maybe the mergeview code has been ported now? https://github.com/codemirror/merge not sure if this is it?

[...]

PR #739, opened 2023-11-20
Fix version parsing

fcollonval:
Follow-up of 4.0.0 release

Xref https://github.com/jupyter/nbdime/issues/659#issuecomment-1819387273

### DIFF
--- a/nbdime/_version.py
+++ b/nbdime/_version.py
@@ -6,7 +6,7 @@ from collections import namedtuple
 
 VersionInfo = namedtuple("VersionInfo", ["major", "minor", "micro", "releaselevel", "serial"])
 
-_specifier_ = {"a": "alpha", "b": "beta", "rc": "candidate", "": "final"}
+_specifier_ = {"a": "alpha", "b": "beta", "rc": "candidate"}
 
 __version__ = "4.0.0"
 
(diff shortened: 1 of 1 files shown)


nbdime/_version.py
```diff
@@ -6,18 +6,19 @@
...
 
 parsed_version = parser.match(__version__)
 groups = parsed_version.groupdict()
+
 version_info = VersionInfo(
     int(groups["major"]),
     int(groups["minor"]),
     int(groups["micro"]),
-    _specifier_[groups.get("releaselevel", "")],
+    _specifier_.get(groups.get("releaselevel", ""), "final"),
```
vidartf, line 22: I'm not sure if this is the right choice: if someone makes a mistake and bumps to "beta" instead of "b", then it will default to final since it cannot parse it?

vidartf: Would you mind sharing what error you were seeing that this will fix?

vidartf:
```suggestion
    _specifier_[groups.get("releaselevel") or ""],
```

fcollonval: The error is that if not found `groups.get("releaselevel") is None`. Therefore, as `None` is not a key of `_specificer`, it raises an `KeyError`.

I'll go for a more advance solution than Vidar's proposal as the propose solution does not address the case of release level being `beta` for example.

fcollonval:
Merging and releasing a patch version

nbdime/_version.py
vidartf: My point was that "beta" would be an invalid value that *should* throw an error. Somewhere at least.. probably not good for it to throw in packaged code though, so this change is fine. But yes, let's try to get a regression test for the future as well.

fcollonval: For documentation, `beta` is a valid specifier: https://peps.python.org/pep-0440/#pre-release-spelling

Issue #468, opened 2019-04-11
Export nbdiff-web html from command line


david-macleod:
Is it possible to export the HTML difference generated by the in-notebook "Export" button directly from the command line (without having to manually interact with the application GUI). If not, would this be difficult to implement as a feature?

The use case is:
1. A large number of notebooks containing plots need to be re-run after a change to the underlying source code is made, and diffs calculated
2. The notebooks which _have_ changed are then inspected visually

If we have a static HTML view of the difference we can automate the process of filtering the notebooks which have not changes. Additionally, having a fixed copy of the HTML difference for future reference is useful it itself.

Thank you

james-nesbitt:
This would also make the diff more usable in various CI scenarios

vidartf:
Sorry, I could swear I wrote an answer to this. It must have gotten lost somewhere..

The current export button was added as a minimal implementation to get this functionality. What you are asking is not currently possible, but should be possible to build. However, it would require one of these two dependencies:
- Either a headless browser to run the JS for generating the HTML DOM from the diff.
- Or node.js with jsdom  package for the same purpose.

Another alternative is to save the diff as a self-contained, dynamic HTML page, i.e.
- The HTML template rendered out, with the diff in JSON format in the page.
- The nbdime JS to convert the diff to DOM (either embedded in the doc, or as a side-by-side .js file).

The final alternative is to rewrite the diff to DOM logic in Python. This is not something that I will do at least, but I'd welcome any PRs, given they come with a commitment to help maintain the code.

I'm good with either of the first two alternatives. Given enough time, I guess both could be implemented, with a flag to choose between them.

Thoughts on the alternatives? (CC @minrk)

minrk:
The second option is probably the simplest to implement and maintain, if not the nicest in principle. Given our current circumstances, I'd probably go with that.

Synergist:
Any update on this issue? My team would greatly benefit from this added functionality.

vidartf:
@Synergist I'm not in a position where I can spend the time on this myself in the near future, but I would be happy to help anyone who would be willing to take this on!

leosunmo:
This would be hugely useful for me as well, unfortunately I am not familiar enough with Python or Jupyter projects to help.

trams:
Is there any progress on this?
If not I will be interested in contributing to this. If someone is willing to pair/review I would appreciate

vidartf:
I'm willing to help with review, and helping out along the way 👍

[...]

PR #744, opened 2024-01-25
Minimal change to permit non-interactive html diff generation (#468)

cjerdonek:
This is a minimal PR to support addressing issue #468. Specifically, it contains a minimal code change needed to let html diffs be created non-interactively (e.g. from the command-line in CI) without having to patch nbdime's source code. A gist demonstrating this (in conjunction with this PR) can be found here: https://gist.github.com/cjerdonek/2dad48eea22846476487d594cd906493

This PR also doesn't even require any changes to the default template, though the gist I wrote above does support providing your own template if you want (which again doesn't require modifying nbdime's source code).

My hope is that, because this PR has a lighter footprint than PR #552, it can provide an interim and more flexible way to address issue #468 in the short-term so people no longer have to patch or fork nbdime to get the benefit of off-line html diffs.

### DIFF
--- a/packages/webapp/src/app/diff.ts
+++ b/packages/webapp/src/app/diff.ts
@@ -244,12 +244,23 @@ function attachToForm() {
  *
  */
 export function initializeDiff() {
-  attachToForm();
-  // If arguments supplied in config, run diff directly:
-  let base = getConfigOption('base');
-  let remote = getConfigOption('remote');
-  if (base && (remote || hasPrefix(base))) {
-    compare(base, remote, 'replace');
+  let diff_data = getConfigOption('diff_data');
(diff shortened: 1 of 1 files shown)


sizhky:
for those who can't wait for the merge and want this functionality, simply run

```
pip install git+https://github.com/cjerdonek/nbdime@issue-468-non-interactive-diffs
```

before trying out the gist provided

cjerdonek:
Thanks, @sizhky. I have no idea why someone "thumbs-downed" this PR when there are no other simple work-arounds available. I've been using this PR with my team for two months now without issues (in conjunction with code similar to what's in the gist).

PR #768, opened 2024-09-05
Update release scripts to user trusted publisher

krassowski:
This should allow releasing the new version again


krassowski:
I configured the repo (environment for the app) but cannot verify nor configure the trusted publishers on the PyPI side. I would need help from @fcollonval or @Carreau or @minrk.

krassowski:
Thank you! It worked, only needed to remember to use `patch` instead of `next` specifier.

PR #779, opened 2025-03-18
cleanup unused mathjax components

bluestealth:
* `MathJaxTypesetter` no longer uses `mathjax_config` or `mathjax_url` after https://github.com/jupyterlab/jupyterlab/pull/13877 so remove `jupyter_server_mathjax` dependency and remove the polyfill
* Update deprecated github actions so CI tests can run


Carreau:
Well, if you agree already, than let's merge.

Issue #785, opened 2025-07-01
Feature: improved and configurable threshold for text similarity


zmbc:
Currently, it appears that if the source of two cells is not at least 70% the same, they will always be treated as separate cells and represented by a delete-cell, then add-cell operation. It feels like this threshold should be much lower, at least if whitespace-only lines are ignored.

There appears to be a relevant TODO:

https://github.com/jupyter/nbdime/blob/458beebb56b9de160abdbb84ae2d19f2d61fbd4e/nbdime/diffing/generic.py#L30-L31

krassowski:
Sure!

zmbc:
@krassowski I've posted a PR for this: #799

PR #799, opened 2026-01-31
Allow configuration of text similarity threshold and improve default

zmbc:
Addresses #785.

Note that I have conservatively made this change _only_ affect the comparison of cell sources. I think you could make an argument that an analogous change for outputs would also be good, but that would have had much larger impacts on the test suite.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -96,8 +96,6 @@ jobs:
             jupyter_server-version: '<2'
           - python-version: '3.11'
             jupyter_server-version: '>=2'
-          - python-version: '3.8'
-            os: windows-latest
           - python-version: '3.11'
             os: windows-latest
 
--- a/docs/source/config.rst
+++ b/docs/source/config.rst
@@ -34,6 +34,8 @@ The current output of `nbdime --config` is:
       Ignore: {}
       attachments: null
       color_words: false
+      text_similarity_ignore_whitespace: true
+      text_similarity_threshold: 0.3
       details: null
       metadata: null
       outputs: null
@@ -45,6 +47,8 @@ The current output of `nbdime --config` is:
       base_url: "/"
       browser: null
       color_words: false
+      text_similarity_ignore_whitespace: true
+      text_similarity_threshold: 0.3
       details: null
       ip: "127.0.0.1"
       metadata: null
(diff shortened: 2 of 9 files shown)


zmbc:
@Carreau I believe I've addressed all your comments!

zmbc:
@krassowski @Carreau what is the next step to move this forward?

Carreau:
Thanks for your patience, unfortunately the blocking part if maintainer time; I've reaproved the workflow; and thanks for the ping. It might take me some time to get back to it as my actual day-time job takes priority and I have few hours to consecrate to open-source outside of that.

zmbc:
@Carreau totally understand. Are you open to onboarding new maintainers?

It looks like this PR isn't working on Python 3.8 which is 1.5 years past EOL. Are you open to a PR to remove 3.8 (and 3.9) support?

Carreau:
Personally yes, but this would have to pass other maintainers as well. But I think it make sens.

Carreau:
limiting to 3.6 was done as of 83a768fa; which was 3 years ago so I think moving to 3.10+ is reasonable.

Carreau:
and https://github.com/jupyter/nbdime/pull/802 has some discussion where @krassowski is ok droping up to 3.8 (included); so 3.9 is on the edge.

zmbc:
Ah, thanks for that context. It looks like 3.8 was still included on Windows in error -- I've removed that now.

zmbc:
The other failing workflow was the "check release" one, with this error:

```
 M package-lock.json

Traceback (most recent call last):
  File "/home/runner/work/nbdime/nbdime/.jupyter_releaser_checkout/scripts/bump_version.py", line 110, in <module>
    bump(args.force, args.spec)
  File "/home/runner/work/nbdime/nbdime/.jupyter_releaser_checkout/scripts/bump_version.py", line 68, in bump
    raise Exception("Must be in a clean git state with no untracked files")
Exception: Must be in a clean git state with no untracked files
```

I'm having a hard time understanding what is going on here, and I'm wondering if it has something to do with the fact that this PR is from a fork.

zmbc:
Want to follow up on my question above:

> Are you open to onboarding new maintainers?

Carreau:
Yes, issue is still time. I just gave you triage right on this repo, this should let you label issues and close stale one.

Carreau:
checks are passing; merging; i suggest you join the jupyter zulip if you wish

PR #808, opened 2026-02-11
Remove unused mock dependency

jamesjer:
All of the tests that use mock import it from unittest.  The old standalone mock module is not needed.


krassowski (approved):
Thanks!

PR #820, opened 2026-08-25
Migrate to jupyter-builder

### LINKED ISSUE #815: Migrate to the new jupyter-builder
I'm in the process of upgrading jupyterlab to 4.6.0 in Fedora Linux, and it seems that extensions need to be migrated as well in order to use the new jupyter-builder properly.

See the migration guide: https://jupyterlab.readthedocs.io/en/latest/extension/extension_migration.html#migrating-to-jupyter-builder

frenzymadness:
Fixes: https://github.com/jupyter/nbdime/issues/815

I'm testing this as an RPM package in Fedora Linux and it builds fine. I'm sorry if this is incomplete, but my knowledge of the JS ecosystem is limited.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["hatchling>=1.5.0", "jupyterlab>=4.0.0,<5"]
+requires = ["hatchling>=1.5.0", "jupyter-builder"]
 build-backend = "hatchling.build"
 
 [project]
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["hatchling>=1.5.0", "jupyterlab>=4.0.0,<5"]
+requires = ["hatchling>=1.5.0", "jupyter-builder"]
```
jtpio, line 2: Wondering if we should keep some bounds on `jupyter-builder`, in case v2 comes with some breaking changes?

For example:


```suggestion
requires = ["hatchling>=1.5.0", "jupyter-builder>=1.2.3,<2"]
```

Here and in other places.

jtpio (commented):
Thanks @frenzymadness for looking into this!

PR #821, opened 2026-08-29
Implement the nbdime:diff command

amannsan:
🤖🤖

Closes #796.

## Summary

- pass `base` and `remote` command arguments to the existing notebook diff widget
- prompt for either notebook path when the caller omits it
- open and optionally activate the resulting diff widget consistently with the other nbdime commands
- add regression coverage for the command-to-widget path

### DIFF
--- a/packages/labextension/src/plugin.ts
+++ b/packages/labextension/src/plugin.ts
@@ -208,8 +208,11 @@ function addCommands(
         name: 'base' | 'remote',
       ): Promise<string | null> {
         const value = args[name];
-        if (typeof value === 'string' && value.trim()) {
-          return value;
+        if (typeof value === 'string') {
+          const path = value.trim();
+          if (path) {
+            return path;
+          }
         }
 
         const label =
@@ -224,9 +227,8 @@ function addCommands(
           placeholder: trans.__('path/to/notebook.ipynb'),
           okLabel: trans.__('Compare'),
         });
-        return result.button.accept && result.value?.trim()
-          ? result.value
-          : null;
+        const path = result.value?.trim();
+        return result.button.accept && path ? path : null;
       }
 
       const base = await getNotebookPath('base');
--- a/packages/labextension/src/plugin.ts
+++ b/packages/labextension/src/plugin.ts
@@ -6,7 +6,7 @@ import type {
   JupyterFrontEnd,
 } from '@jupyterlab/application';
 
-import { CommandToolbarButton } from '@jupyterlab/apputils';
+import { CommandToolbarButton, InputDialog } from '@jupyterlab/apputils';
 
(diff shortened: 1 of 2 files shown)


packages/labextension/src/plugin.ts
```diff
@@ -198,11 +203,54 @@ function addCommands(
...
-      // if missing, prompt with dialog.
-      //let content = current.notebook;
-      //diffNotebook({base, remote, translator});
+    execute: async args => {
+      async function getNotebookPath(
+        name: 'base' | 'remote',
+      ): Promise<string | null> {
+        const value = args[name];
+        if (typeof value === 'string' && value.trim()) {
+          return value;
```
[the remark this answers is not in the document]
amannsan: Fixed in be21279: both command arguments and dialog values are now trimmed before the diff request, and the regression test covers whitespace-padded paths.

packages/labextension/src/plugin.ts
```diff
@@ -198,11 +203,56 @@ function addCommands(
   });
 
   commands.addCommand(CommandIDs.diffNotebook, {
-    execute: args => {
-      // TODO: Check args for base/remote
-      // if missing, prompt with dialog.
-      //let content = current.notebook;
-      //diffNotebook({base, remote, translator});
+    execute: async args => {
```
krassowski, line 231: We would also want to add description of arguments: https://eslint-plugin.readthedocs.io/en/latest/rules/command-described-by/

krassowski (commented):
In the future I would like this command to also support accepting notebook json as in https://github.com/jupyterlab/jupyter-collaboration/pull/589. It is worth thinking what shape the arguments should have to allow passing either path or source JSON.

amannsan:
Thanks — I agree that the command arguments should leave room for in-memory notebooks.

The smallest stable shape seems to be keeping the `base` and `remote` keys and allowing each value to be either a path string or an `nbformat.INotebookContent` object. The string/object distinction is unambiguous, and callers would not need different command keys when the in-memory path is added. This PR implements only the string branch because `diffNotebook()` currently uses the server-backed path API; computing and rendering an in-memory diff is separate work.

The label check is also red because the PR has no triage label. My repository permission is read-only, so I cannot add the existing `bug` label from issue #796.

packages/labextension/src/plugin.ts
amannsan: Added in 35072fb. The command now declares a `describedBy` schema for every argument it currently consumes: optional `base` and `remote` path strings, plus optional `activate`. The path arguments remain optional because the command opens the corresponding dialog when either one is omitted.

The schema describes the implemented string-path contract only; it can be widened alongside the future in-memory notebook implementation.

The focused Jest test, full TypeScript build, and Prettier check all pass.

